Issue #34, opened 2018-11-21
[Clarification] Size of sentence vectors


Colanim:
From `README.md` :

> Each sentence is translated to a 768-dimensional vector. One exception is REDUCE_MEAN_MAX pooling strategy, which translates a sentence into a 1536-dimensional vector.

**Why the sentence vector's size does not change with the number of layers chosen ?**

---

From `README.md` :

> pooling_layer : the encoding layer that pooling operates on, where -1 means the last layer, -2 means the second-to-last, etc.

If `-pooling_layer=-4`, I expected to have 4 vectors of size 768 concatenated into 1 vectors of size 4 * 768 = 3072, because in the BERT [paper](https://arxiv.org/pdf/1810.04805.pdf) :

> The best performing method is to concatenate the token representations from the top four hidden layers of the pre-trained Transformer

hanxiao:
in the current version, `-pooling_layer=-4` means fourth-to-last, namely the 9th layer of a 12-layer BERT or the 21st layer of a 24-layer BERT.

i will allow `-pooling_layer` to be a list in the future, so you can use it as `-pooling_layer=[-4,-3,-2,-1]`.

Colanim:
Is there any advantage of using something else than the last layer ? Concatenating several last layers increase accuracy (from the BERT paper), but taking a single layer other than the last layer will decrease accuracy, right ?

hanxiao:
Second-to-last layer is less biased towards the training targets (i.e. masked LM and next sentence prediction), it's also what the BERT author suggested:
https://github.com/google-research/bert/issues/71#issuecomment-436507081

Colanim:
#55

dhruvsakalley:
Hi Han,
I was wondering what dimension sentence embedding should I expect with -pooling_layer -4 -3 -2 -1 and -pooling_strategy=REDUCE_MEAN ? currently I'm getting 4096, instead of 1024 from a large model, would it be alright averaging them after splitting the 4 layers at 1024?

PR #55, opened 2018-11-27
Multilayers support added

Colanim:
Implementation of multi-layers added : 

It's now possible to specify several layers, and the resulting feature vector will be the concatenation of these layers.

From #34


hanxiao:
awesome! thanks for your contribution!

Issue #19, opened 2018-11-16
Classifier Predictions


loretoparisi:
Thanks for this service, it works like a charm and better than Google's!
How it difficult to support classifier predictions for the classification task [here](https://github.com/google-research/bert#prediction-from-classifier) from a pre-trained model?

hanxiao:
Thanks! Supporting classification isn't hard. I will implement it as a feature in the future.

For now, I'd like to implement more pooling strategies.

loretoparisi:
@hanxiao thank you that would be awesome! I was looking at the bert repo classification task here https://github.com/google-research/bert/blob/60454702590a6c69bd45c5d4258c7e17b8a3e1da/run_classifier.py
By the way this is far more complex  how it could actually be.

mortimer125:
hi, could I use this sentence vector with pooling_strategy being CLS_TOKEN to do classification tasks such as relation extraction ?

hanxiao:
@loretoparisi I write one based on `tf.estimator.DNNClassifier`, checkout [example5](example5.py) is this something you have in mind?

loretoparisi:
@hanxiao amazing 💯 🥇 This is exactly what I was looking for! The reference [CAIL](https://github.com/thunlp/CAIL/blob/master/README_en.md#221-the-fields-of-data) dataset is also a very good example for both the text embedding and labels encoding.
Closing! Thank you!

hanxiao:
FYI, I made a visualization of different BERT layers, may help you understand the pool_layer and pool_strategy. See: https://github.com/hanxiao/bert-as-service#q-so-which-layer-and-which-pooling-strategy-is-the-best

PR #63, opened 2018-11-28
Add classification example

hanxiao:


hanxiao:
[example5.py](example5.py) should fix #19  in some way, at least

Issue #39, opened 2018-11-22
On Windows, I got zmq.error:ZMQError: Protocal not supported when executing 'python app.py'


BruceDai003:
Your project is awesome. But I'm not sure if it will work on Windows 10 platform. I just cloned your project, and downloaded the BERT pre-trained model. The moment I run `python app.py -model_dir F:\data\chinese_L-12_H-768_A-12\chinese_L-12_H-768_A-12/ -num_worker=4` , I got an error:
```
λ python app.py -model_dir F:\data\chinese_L-12_H-768_A-12\chinese_L-12_H-768_A-12/ -num_worker=4
usage: app.py -model_dir F:\data\chinese_L-12_H-768_A-12\chinese_L-12_H-768_A-12/ -num_worker=4
                 ARG   VALUE
__________________________________________________
      max_batch_size = 256
         max_seq_len = 25
           model_dir = F:\data\chinese_L-12_H-768_A-12\chinese_L-12_H-768_A-12/
          num_worker = 4
       pooling_layer = -2
    pooling_strategy = REDUCE_MEAN
                port = 5555

Exception in thread Thread-1:
Traceback (most recent call last):
  File "D:\Anaconda3\lib\threading.py", line 916, in _bootstrap_inner
    self.run()
  File "F:\Work\Github\bert-as-service\service\server.py", line 72, in run
    self.backend.bind('ipc://*')
  File "zmq/backend/cython/socket.pyx", line 495, in zmq.backend.cython.socket.Socket.bind (zmq\backend\cython\socket.c:5653)
  File "zmq/backend/cython/checkrc.pxd", line 25, in zmq.backend.cython.checkrc._check_rc (zmq\backend\cython\socket.c:10014)
zmq.error.ZMQError: Protocol not supported
```

I have no idea what this zmq is, and I googled, it seems that 'ipc' is not supported on Windows, we should use 'tcp' instead. I tried to just change 'ipc' to 'tcp' on line 72, but still got the similar error:
```
λ python app.py -model_dir F:\data\chinese_L-12_H-768_A-12\chinese_L-12_H-768_A-12/ -num_worker=4
usage: app.py -model_dir F:\data\chinese_L-12_H-768_A-12\chinese_L-12_H-768_A-12/ -num_worker=4
                 ARG   VALUE
__________________________________________________
      max_batch_size = 256
         max_seq_len = 25
[...]

PR #96, opened 2018-12-06
Release to pypi

### LINKED ISSUE #50: publish to pypi
It would be nice to be able to depend on this library directly from [pipy](https://pypi.org).
Ideally, there would be two packages published: one for the client (without a dependency to tensorflow) and one for the server.

hanxiao:
- fix #50


hanxiao:
add windows support by switching to tcp and bind port randomly, fix #39 
the package is now available on pypi, fix #50

PR #101, opened 2018-12-07
Fix serious bug in re-sort algorithm.

likejazz:
It treats a byte as a string by default. So, if you sort `1, 5, 10`, it will return the wrong result as `1, 10, 5` instead of `1, 5, 10`. Therefore, It's necessary to explicitly convert byte to int.


hanxiao:
fyi it's now available on pypi 1.4.7

PR #103, opened 2018-12-07
Fix docker

### LINKED ISSUE #98: Dockerfile need to be updated
Dockerfile needs to be updated, and there is no client version 1.4.6 (python2.x).

I tried to fix it, the following can be used.

```
FROM tensorflow/tensorflow:1.12.0-gpu-py3
COPY ./ /app
COPY ./mydocker/entrypoint.sh /app
WORKDIR /app
RUN pip install -r requirements.gpu.txt
RUN pip install bert-serving-server==1.4.5
#RUN python /app/server/setup.py install
ENTRYPOINT ["/app/entrypoint.sh"]
CMD []
```

hanxiao:
fix #98

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ One may also run BERT Service in a container:
 ```bash
 docker build -t bert-as-service -f ./docker/Dockerfile .
 NUM_WORKER=1
-PATH_MODEL=<path of your model>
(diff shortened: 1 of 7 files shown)


hanxiao:
add close() method to client

PR #118, opened 2018-12-12
move tmp files which are generated by ipc socket to tmp folder

yyq263:
try to solve issue in https://github.com/hanxiao/bert-as-service/issues/73


hanxiao:
solved in #127 , close

Issue #73, opened 2018-11-30
[Enhancement] Choose the location where to create tmp files


Colanim:
Where I run the server side, I have plenty of `tmp` files accumulating. 

**Maybe add a server option to specify a temporary folder location.**

_(I'll look into it when I have time)_

PR #127, opened 2018-12-13
Tmp file (generated by ipc socket) location choice 

Colanim:
**Code greatly inspired by #118** 

It's basically the same pull request, but you can choose the location of file using an environment variable. The default behavior (the environment variable is not set) is the same as previously : tmp file are created on the current folder.

**Solve #73** (I was indeed talking about socket tmp files generated ^^)

_Note : the choice of the location is done through environment variable. In my opinion, it's less invasive than command line argument._

### DIFF
--- a/README.md
+++ b/README.md
@@ -389,6 +389,11 @@ bc2 = BertClient()
 bc2.encode(lst_str)
 ```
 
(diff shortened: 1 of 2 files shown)


hanxiao:
looks good to me, thanks!

Issue #120, opened 2018-12-12
How I serve a fine-tuned model?


rxy1212:
**Prerequisites**

> Please fill in by replacing `[ ]` with `[x]`.

* [x ] Are you running the latest `bert-as-service`?
* [x ] Did you follow [the installation](https://github.com/hanxiao/bert-as-service#install) and [the usage](https://github.com/hanxiao/bert-as-service#usage) instructions in `README.md`?
* [x ] Did you check the [FAQ list in `README.md`](https://github.com/hanxiao/bert-as-service#faq)?
* [x ] Did you perform [a cursory search on existing issues](https://github.com/hanxiao/bert-as-service/issues)?

**System information**

> Some of this information can be collected via [this script](https://github.com/tensorflow/tensorflow/tree/master/tools/tf_env_collect.sh).

- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): linux server
- TensorFlow installed from (source or binary): binary
- TensorFlow version: 1.11.0 cuda9 cudnn7
- Python version: 3.6.5
- `bert-as-service` version: 1.5.0
- GPU model and memory: T100 16G * 2
- CPU model and memory: 

---
I just fine-tuned a bert model for sentence similarity task. But when I serve my model by using bert-as-service, it still output a vector of 768 dims. Now I want to start a service which input two sentences and output two float number ["probability of 0", "probability of 1"] like set "do_predict=true" when  running run_classifier.py. It's any advice for serving a fine-tuned model?

...

Colanim:
Just implement your code (add the classification layer) into Bert-as-service.
If you don't want to touch Bert-as-service, you have to implement your classification layer outside of Bert-as-service.

Issue #76, opened 2018-11-30
Finetuning Example


simonefrancia:
Hi, 
I am trying your code example (example5.py) to make a fine-tuning on my own mood dataset for a text classification task. I have already adapted the code and train starts correctly. 
After 5000 steps loss is descending but accuracy on validation is poor and constant at 42%. 

My doubt is that the fine-tuning is modifying weights of prediction task but not the input embeddings. When I used ELMo embedding for example, I set that embedding of input were trainable and classification task results are really better than this. 

Any suggestion about that? 

Thanks

hanxiao:
my bad. I forgot to mask the sequence when doing reduce_mean, reduce_max and reduce_mean_max. this is fixed in #94 and now merged into master. please give it a try and may produce different result.

simonefrancia:
I have tried new version but my training doesn't seem to be better. 
```
INFO:tensorflow:Calling model_fn.
INFO:tensorflow:Done calling model_fn.
INFO:tensorflow:Create CheckpointSaverHook.
INFO:tensorflow:Graph was finalized.
INFO:tensorflow:Running local_init_op.
INFO:tensorflow:Done running local_init_op.
INFO:tensorflow:Saving checkpoints for 0 into data/model/1206-081741/model.ckpt.
INFO:tensorflow:loss = 264.92365, step = 1
INFO:tensorflow:global_step/sec: 0.194579
INFO:tensorflow:loss = 142.13977, step = 101 (513.941 sec)
INFO:tensorflow:global_step/sec: 0.194874
INFO:tensorflow:loss = 152.33533, step = 201 (513.145 sec)
INFO:tensorflow:global_step/sec: 0.19485
INFO:tensorflow:loss = 159.28421, step = 301 (513.216 sec)
INFO:tensorflow:global_step/sec: 0.194599
INFO:tensorflow:loss = 152.00076, step = 401 (513.898 sec)
INFO:tensorflow:global_step/sec: 0.194962
INFO:tensorflow:loss = 144.81601, step = 501 (512.911 sec)
INFO:tensorflow:global_step/sec: 0.194442
INFO:tensorflow:loss = 134.33696, step = 601 (514.280 sec)
INFO:tensorflow:global_step/sec: 0.192623
INFO:tensorflow:loss = 144.52493, step = 701 (519.150 sec)
INFO:tensorflow:global_step/sec: 0.193265
INFO:tensorflow:loss = 144.21286, step = 801 (517.422 sec)
[...]

PR #147, opened 2018-12-19
add fine-tuned model serving to CLI

hanxiao:
- serving fine-tuned model in a more user-friendly way


hanxiao:
fix #120 
related to #76

PR #177, opened 2019-01-07
For Client: update note and set zmq.LINGER zero

chenyou-fdu:
* change note: only set `check_version=False` will still send message to server, need set `check_length=False`.
* call BertClient.close() will hang at `self.context.term()`, if server_status is quried from server and timeout. Refer to [StackoverFlow #1](https://stackoverflow.com/questions/46029965/zeromq-blocked-in-a-context-term-call-why-how-to-prevent) and [StackoverFlow #2](https://stackoverflow.com/questions/44273941/python-script-not-terminating-after-timeout-in-zmq-recv), I set zmq.LINGER to zero, and it worked. But I'm not sure is this option will lead to any other problem, so pls help to review, thx.

### DIFF
--- a/client/bert_serving/client/__init__.py
+++ b/client/bert_serving/client/__init__.py
@@ -41,7 +41,7 @@ class BertClient:
 
         Create a BertClient that connects to a BertServer.
         Note, server must be ready at the moment you are calling this function.
-        If you are not sure whether the server is ready, then please set `check_version=False`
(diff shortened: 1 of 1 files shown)


hanxiao:
thanks a lot. LGTM.

hanxiao:
maybe it's better to move set linger to `BertClient.close()`

Issue #185, opened 2019-01-11
benchmark execution error


qnamqj:
### 1、the first error is:
python benchmark.py 

Traceback (most recent call last):
  File "benchmark.py", line 84, in <module>
    server = BertServer(args)
  File "/miniconda3/envs/bert/lib/python3.6/site-packages/bert_serving/server/__init__.py", line 43, in __init__
    self.logger = set_logger(colored('VENTILATOR', 'magenta'), args.verbose)
AttributeError: type object 'args_nt' has no attribute 'verbose'

### 2、the second error is:
then i read the source, find 
  server = BertServer(args)
because some default property in helper.py that not constain in the args, 
so i add this property to the benchmark.py's common object

common = {
    ......

    'verbose': False,
    'tuned_model_dir': None,
    'ckpt_name': 'bert_model.ckpt',
    'config_name': 'bert_config.json',
    'mask_cls_sep': False,
    'priority_batch_size': 16,
    'device_map': []
}

then again exection  `python benchmark.py`, bet get the new error:

I:VENTILATOR:[__i:__i: 62]:freeze, optimize and export graph, could take a while...
I:GRAPHOPT:[gra:opt: 48]:model config: /chinese_L-12_H-768_A-12/bert_config.json
I:GRAPHOPT:[gra:opt: 50]:checkpoint: /chinese_L-12_H-768_A-12/bert_model.ckpt
I:GRAPHOPT:[gra:opt: 54]:build graph...
I:GRAPHOPT:[gra:opt:121]:load parameters from checkpoint...
I:GRAPHOPT:[gra:opt:123]:freeze...
I:GRAPHOPT:[gra:opt:126]:optimize...
I:GRAPHOPT:[gra:opt:134]:write graph to a tmp file: /tmp/tmpf6x979_k
I:VENTILATOR:[__i:__i: 70]:optimized graph is stored at: /tmp/tmpf6x979_k
I:VENTILATOR:[__i:_ru:102]:bind all sockets
I:VENTILATOR:[__i:_ru:106]:open 8 ventilator-worker sockets
I:VENTILATOR:[__i:_ru:109]:start the sink
I:SINK:[__i:_ru:254]:ready
I:VENTILATOR:[__i:_ge:177]:get devices
I:VENTILATOR:[__i:_ge:209]:device map: 
    worker  0 -> gpu  0
    worker  1 -> gpu  1
[...]

PR #188, opened 2019-01-14
move zmq linger to initialization

### LINKED ISSUE #187: Bert Client: ZeroMQ Socket lingering issue after timeout.
**Prerequisites**

> Please fill in by replacing `[ ]` with `[x]`.

* [x] Are you running the latest `bert-as-service`?
* [x] Did you follow [the installation](https://github.com/hanxiao/bert-as-service#install) and [the usage](https://github.com/hanxiao/bert-as-service#usage) instructions in `README.md`?
* [x] Did you check the [FAQ list in `README.md`](https://github.com/hanxiao/bert-as-service#speech_balloon-faq)?
* [x] Did you perform [a cursory search on existing issues](https://github.com/hanxiao/bert-as-service/issues)?

**System information**

- OS Platform and Distribution (e.g., Linux Ubuntu 16.04):
- TensorFlow installed from (source or binary):
- TensorFlow version: 1.4.0
- Python version: 3.5.2
- `bert-as-service` version: 1.6.9

---

When I run this script with no active server:

```python
import sys
from bert_serving.client import BertClient

try:
    bert_client = BertClient(timeout=500)
except TimeoutError as e:
    print(e)
sys.exit(0)
```

A time out error is raised, then the script refuses to exit even if sys.exit() is called.

After a brief search on Stack Overflow, I found the following questions:

https://stackoverflow.com/questions/44273941/python-script-not-terminating-after-timeout-in-zmq-recv
https://stackoverflow.com/questions/17140417/termination-of-python-script-while-using-zeromq-with-dead-server

The default behavior of ZMQ is to keep the socket instance alive indefinitely even after timeout. The 'dead' socket instance blocks the script from exiting in the above example. I am not sure if this is going to cause other side-effects in a different setting, but I believe this should be fixed.

Probably we should have LINGER = 0 or some small number (like 200-ish) ?

hanxiao:
fix #187

### DIFF
--- a/benchmark.py
+++ b/benchmark.py
@@ -58,7 +58,7 @@ class BenchmarkClient(threading.Thread):
 
     def run(self):
         time_all = []
-        bc = BertClient(port=PORT, port_out=PORT_OUT, show_server_config=False, check_version=False)
+        bc = BertClient(port=PORT, port_out=PORT_OUT, show_server_config=False, check_version=False, check_length=False)
         for _ in range(self.num_repeat):
             start_t = time.perf_counter()
             bc.encode(self.batch)
--- a/benchmark.py
+++ b/benchmark.py
@@ -34,6 +34,7 @@ common = {
     'ckpt_name': 'bert_model.ckpt',
     'tuned_model_dir': None,
     'mask_cls_sep': False,
+    'device_map': []
 }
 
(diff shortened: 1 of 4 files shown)


hanxiao:
fix #185

PR #202, opened 2019-01-18
Add fp16 support

hanxiao:


hanxiao:
It works. but given that limited support from GPU, not sure how useful this will be.

hanxiao:
Fix fp16 support mentioned in https://github.com/google-research/bert/issues/249

Issue #194, opened 2019-01-16
'bert-serving-start' is not recognized as an internal or external command


sailormoon2016:
Hi, 
This is a very silly question.....
I have python 3.6.6, tensorflow 1.12.0, doing everything in conda environment, Windows 10.
I pip installed bert-serving-server/client and it shows 
`Successfully installed GPUtil-1.4.0 bert-serving-client-1.7.2 bert-serving-server-1.7.2 pyzmq-17.1.2`
but when I run the following as CLI
`bert-serving-start -model_dir /tmp/english_L-12_H-768_A-12/ -num_worker=4`
it says
`'bert-serving-start' is not recognized as an internal or external command`

I found bert-serving library is located under C:\Users\Name\Anaconda\Lib\site-packages. So I tried to run bert-serving-start again under these three folders:
1. site-packages
2. site-packages\bert_serving
3. site-packages\bert_serving_server-1.7.2.dist-info

![](http://i67.tinypic.com/4ke788.png)
However, the result is same as not recognized. Can anyone help me?

Colanim:
Did you try adding `site-packages\bert_serving` to your `PATH` variable ?

hanxiao:
@sailormoon2016 please refer to the step 1 of https://github.com/hanxiao/bert-as-service/issues/99#issuecomment-452769280

for the latest `bert-as-service` >=1.7.0, you don't need to do step 2 in that comment

hanxiao:
fyi, this is fixed in #212, the new feature is available since 1.7.7. Please do 
```bash
pip install -U bert-serving-server bert-serving-client
```
for the update.

A6Matrix:
> Hi,
> This is a very silly question.....
> I have python 3.6.6, tensorflow 1.12.0, doing everything in conda environment, Windows 10.
> I pip installed bert-serving-server/client and it shows
> `Successfully installed GPUtil-1.4.0 bert-serving-client-1.7.2 bert-serving-server-1.7.2 pyzmq-17.1.2`
> but when I run the following as CLI
> `bert-serving-start -model_dir /tmp/english_L-12_H-768_A-12/ -num_worker=4`
> it says
> `'bert-serving-start' is not recognized as an internal or external command`
> 
> I found bert-serving library is located under C:\Users\Name\Anaconda\Lib\site-packages. So I tried to run bert-serving-start again under these three folders:
> 
> 1. site-packages
> 2. site-packages\bert_serving
[...]

Issue #204, opened 2019-01-19
FP16 speed benchmark


hanxiao:
https://github.com/google-research/bert/issues/378

Since [1.7.5 of `bert-as-service`](https://github.com/hanxiao/bert-as-service/) a new option `-fp16` is added to the CLI, which loads a pretrained/fine-tuned BERT model (originally in FP32) and converts it to a FP16 frozen graph. As a result the model size and the memory footprint reduces 40% comparing to the FP32 counterpart. On the other hand, the speedup highly depends on the device, GPU driver, CUDA, etc. I currently don't have such environment to measure the speedup with `-fp16`.

If someone has a half-precision enabled device, it would be really awesome if you can test the performance and report it here or in [`bert-as-service` issues](https://github.com/hanxiao/bert-as-service/issues). Thanks in advance! 🙇

davidlenz:
Would love to help. I am running a Titan V on Win 10 with the following specs

![grafik](https://user-images.githubusercontent.com/25869733/51473423-9c7b7e80-1d7c-11e9-875a-fbdb79044857.png)


I guess i just run `benchmark.py` and add the `-fp16` flag to line 16, i.e. 

`common = vars(get_args_parser().parse_args(['-model_dir', MODEL_DIR, '-port', str(PORT), '-port_out', str(PORT_OUT)]))`, 

is that correct?

Also, should the benchmark be run with the chinese language model as in `benchmark.py` or can i stick with the multi language model i am currently running?

hanxiao:
Thanks a lot for your initiative! Yes, correct. Just add `-fp16` to 
https://github.com/hanxiao/bert-as-service/blob/6d57dddf2528751cfd8e8ea470c22df4e8651362/benchmark.py#L16
and it should work.

No, you don't need Chinese BERT at all. You can use whatever BERT you like, as the speed is measured on the sample-level not on the token-level, changing BERT language should not affect the speed.

davidlenz:
Thanks for the clarifications! The benchmark is currently running, however i had to do some additional steps, which i will shortly outline in the following, to determine whether i introduced incosistencies.

### Why not just run benchmark.py as is?
Since bert-server-start does not work for me as described in #99, i used the fix suggested by https://github.com/hanxiao/bert-as-service/issues/99#issuecomment-452769280 , i.e. starting the server via `python start-bert-as-service.py` with the necessary flags. 

### So why is this a problem for benchmark.py?
[...]

PR #212, opened 2019-01-22
refactor benchmark add fp16 result

hanxiao:


hanxiao:
fix #194 on Windows

hanxiao:
refactor benchmark for better evaluation #204

hanxiao:
fp16 will be added in a separate PR.

PR #265, opened 2019-03-12
Simple dot products aren't sensible scoring functions

GabrielBianconi:
The embeddings generated by this library have different norms depending on the input. In my understanding, using a simple dot product as a scoring function for search isn't sensible because the norm of the vector affects the output irrespective of relatedness. 

Cosine similarity is a more suitable metric. But, since the query is fixed across all comparisons, we can safely ignore the query's norm. Thus, a simple solution is to normalize the dot products with the norm of the document.

Qualitatively, the results I get from simple dot products are pretty bad, whereas they're as expected with the normalized version.


hanxiao:
👍

PR #308, opened 2019-04-04
Add request verification to avoid crash

JTiger0431:
When server ports are exposed to the public network, there is a risk of malicious requests or scans. The server will crash when the number of frames is satisfied but the last two frames are not integer. So, I added two assert statements to avoid this problem.

### DIFF
--- a/server/bert_serving/server/__init__.py
+++ b/server/bert_serving/server/__init__.py
@@ -166,7 +166,9 @@ class BertServer(threading.Thread):
             try:
                 request = frontend.recv_multipart()
                 client, msg, req_id, msg_len = request
-            except ValueError:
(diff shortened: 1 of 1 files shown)


hanxiao:
this unfortunately will always fail, as `req_id` and `msg_len` are bytes(-string) not digits.

JTiger0431:
I know. That's why I used this expression. I have already tested it via bert-serving-client-1.8.9.
Refer to this: https://docs.python.org/3/library/stdtypes.html#str.isdigit

hanxiao:
Thanks a lot! But notice that exposing this service directly to the public internet is not recommended. Securing the incoming traffic (e.g. via encryption) is out of scope of this project.

PR #341, opened 2019-05-07
fix example7

paperplanet:
Fix example7 (PCA visualization) error related to namedtuple. Tested with bert-as-service v1.8.9.


hanxiao:
thanks a lot! 🙇

paperplanet:
My pleasure!

Issue #400, opened 2019-07-04
Disastrous performance downgrade after 30mins of continuous requests


NicksonSone:
**Prerequisites**

> Please fill in by replacing `[ ]` with `[x]`.

* [ ] Are you running the latest `bert-as-service`?
* [ ] Did you follow [the installation](https://github.com/hanxiao/bert-as-service#install) and [the usage](https://github.com/hanxiao/bert-as-service#usage) instructions in `README.md`?
* [x] Did you check the [FAQ list in `README.md`](https://github.com/hanxiao/bert-as-service#speech_balloon-faq)?
* [x ] Did you perform [a cursory search on existing issues](https://github.com/hanxiao/bert-as-service/issues)?

**System information**
> Some of this information can be collected via [this script](https://github.com/tensorflow/tensorflow/tree/master/tools/tf_env_collect.sh).

- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Centos 7
- TensorFlow installed from (source or binary): binary
- TensorFlow version:1.12
- Python version: 3.6
- `bert-as-service` version:  server 1.83 client 1.91
- GPU model and memory: Tesla P100 16G
- CPU model and memory:16 Cores, 118G

---

### Description
I've 100% followed homepage tutorial to get sentence embedding. Everything works fine until BertServer had processed billions of requests from BertClient .Then, Server side suffered severe speed degradation from taking just 200ms to encode a batch of 200 sentences to 10s. I personally make sure a new BertClient is created for each request under multiprocessing circumstance.

Then I try to locate which part of code has taken so much time. It turn out the following snippet in _run method of BertServer can take a few hundred microseconds or even 10 seconds to execute, depending on the number of requests server has satisfied. This dictionary actually contains statistics about server status, like number of requests received from startup. Every time a BertClient is created and send a request, the following code is executed and BertClient get status info about server side.

Actually, I created a pool of 30 processes and continuously send the same request of batch 200 to server side for, like, 30mins or even more. I've also add a few `logger.info` between each possible line of code causing my issue. Of course, logger is configured to log time, microsecond precision. In my code, there are two logger.info before and after the following snippet.

`
                    status_runtime = {'client': client.decode('ascii'),
                                      'num_process': len(self.processes),
                                      'ventilator -> worker': addr_backend_list,
                                      'worker -> sink': addr_sink,
                                      'ventilator <-> sink': addr_front2sink,
                                      'server_current_time': str(datetime.now()),
[...]

PR #401, opened 2019-07-05
Improve statistics performance

BillDorn:
I think this will fix issue #400.

By default creating a new client requests server config to check versions, this also builds and sends the server statistics which the client does not use. I've broken it up into `SHOW_CONFIG` and `SHOW_STATUS` so statistic object will only be created/sent when you explicitly request it from client/http endpoint.

I believe the client keyed statistic dicts continue to grow unbounded as long as new clients are created using up memory and slowing down code path. I've replaced them with a dict-like class that takes a set size limit. When the capacity is reached it evicts entries using LRU policy; evicted entries are added to aggregate stats fields for generating info.
These stats may be slightly off if you use more than the capacity of clients simultaneously, but the limit is currently 500 which seems pretty high.

I've changed the calculation for `size_per_request`, the map `{1:100, 2:1}` should have an average size per req of 102/101 not 3/2.

### DIFF
--- a/client/bert_serving/client/__init__.py
+++ b/client/bert_serving/client/__init__.py
@@ -105,7 +105,7 @@ class BertClient(object):
         self.token_info_available = False
 
         if not ignore_all_checks and (check_version or show_server_config or check_length or check_token_info):
-            s_status = self.server_status
+            s_status = self.server_config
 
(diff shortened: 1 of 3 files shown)


NicksonSone:
Thanks for the effort.  really appreciated

hanxiao:
wow thanks a lot!

PR #421, opened 2019-07-21
Updating bert_serving_start to bert-serving-start

vsoch:
This is a quick pull request to change instances of `bert_serving_start` to `bert-serving-start`, which seems to be the correct entrypoint for the module!

```python
        'console_scripts': ['bert-serving-start=bert_serving.server.cli:main',
```

I would bet it used to be the other one? And then someone decided that dashes > underscore :)

Signed-off-by: Vanessa Sochat <vsochat@stanford.edu>


hanxiao:
aha, you are totally right! appreciated, thanks a lot 🙇

Issue #449, opened 2019-09-19
Long Sequence AssertionError


SimulatedANeal:
**Prerequisites**

> Please fill in by replacing `[ ]` with `[x]`.

* [x ] Are you running the latest `bert-as-service`?
* [x ] Did you follow [the installation](https://github.com/hanxiao/bert-as-service#install) and [the usage](https://github.com/hanxiao/bert-as-service#usage) instructions in `README.md`?
* [x ] Did you check the [FAQ list in `README.md`](https://github.com/hanxiao/bert-as-service#speech_balloon-faq)?
* [x ] Did you perform [a cursory search on existing issues](https://github.com/hanxiao/bert-as-service/issues)?

**System information**

> Some of this information can be collected via [this script](https://github.com/tensorflow/tensorflow/tree/master/tools/tf_env_collect.sh).

- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): CentOS Linux
- TensorFlow installed from (source or binary): source
- TensorFlow version: 1.14.0 (GPU)
- Python version: 3.6.6
- `bert-as-service` version: 1.9.7
- GPU model and memory: Tesla T4, 16GB

---

### Description

For very long (> 512 tokens) previously un-tokenized input sequences, I am getting an AssertionError:

```
File "/home/ndigre/.virtualenvs/eb0.6.0/lib64/python3.6/site-packages/bert_serving/server/bert/extract_features.py", line 116, in convert_lst_to_features
    assert len(input_ids) == max_seq_length
```

I think this might have something to do with the changes made in [#447]. Specifically, it seems that the conditional check `... and (no_special_token and is_tokenized):` is preventing the sequence from being truncated. Is this the intended behavior?

I'm using this command to start the server:

```python
args = [
    "-model_dir", "multi_cased_L-12_H-768_A-12",
    "-num_worker=1",
    "-max_seq_len='NONE'",
    "-max_batch_size=32",
    "-port={}".format(port),
    "-port_out={}".format(port_out),
    "-device_map=0",
    "-cased_tokenization",
]
server = BertServer(args)
server.start()
```

and calling the server via:
```python
bc = BertClient(port=port, port_out=port_out)
bc.encode(long_texts)
```

OnionWang:
I met this problem too, I think the author missed a 'not' in condition, it should be “if len(tokens_a) > max_seq_length - 2 and **not** (no_special_token and is_tokenized)”

hanxiao:
[...]

Issue #448, opened 2019-09-18
Server crash if i use new version server (1.9.7)


Mekkeliard:
**Prerequisites**

> Please fill in by replacing `[ ]` with `[x]`.

* [x] Are you running the latest `bert-as-service`?
* [x] Did you follow [the installation](https://github.com/hanxiao/bert-as-service#install) and [the usage](https://github.com/hanxiao/bert-as-service#usage) instructions in `README.md`?
* [x] Did you check the [FAQ list in `README.md`](https://github.com/hanxiao/bert-as-service#speech_balloon-faq)?
* [x] Did you perform [a cursory search on existing issues](https://github.com/hanxiao/bert-as-service/issues)?

**System information**

> Some of this information can be collected via [this script](https://github.com/tensorflow/tensorflow/tree/master/tools/tf_env_collect.sh).

- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Windows 10 Pro (on docker)
- TensorFlow installed from (source or binary): binary
- TensorFlow version: 1.14.0
- Python version: 3.7.4
- `bert-as-service` version:  1.9.7
- GPU model and memory: NVidia GTX-650ti
- CPU model and memory: Intel Core i5 3330

---

### Description

I'm using this command to start the server:

```bash
bert-serving-start -num_worker=1 -http_port 80 -model_dir /model
```

I'm use REST Api (or ZeroMq - i checked in two versions) and trace (from C# client) phrase on russian language: "Спасибо))) Я просто проверял возможность на элементарные ответы))) Надеюсь Иру доделают до совершенства)".
If i use version 1.9.6, then everything works well. 
But if i use version server 1.9.7, then server crash with error.

Please tell me why this is happening?

Then this issue shows up:

```
Process BertWorker-3:
Traceback (most recent call last):
File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/client/session.py", line 1356, in _do_call
return fn(*args)
File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/client/session.py", line 1341, in _run_fn
options, feed_dict, fetch_list, target_list, run_metadata)
File "/usr/local/lib/python3.6/dist-packages/tensorflow/python/client/session.py", line 1429, in _call_tf_sessionrun
run_metadata)
tensorflow.python.framework.errors_impl.UnknownError: AssertionError:
Traceback (most recent call last):

[...]

PR #450, opened 2019-09-20
bugfix for #449

yangyaofei:


hanxiao:
close #449 #448 
fix is rolled out in 1.9.8

Issue #394, opened 2019-06-27
Error with Stop BertServer from Python


lu161513:
My Code:

`from bert_serving.server.helper import get_args_parser`
`from bert_serving.server import BertServer`
`from bert_serving.client import BertClient`

`args = get_args_parser().parse_args(['-model_dir', '/home/cn-pvgtoddlu/model/chinese_L-12_H-768_A-12',
                                     '-num_worker','1',
                                     '-port', '5555',
                                     '-port_out', '5556',
                                     '-max_seq_len', 'NONE',
                                     '-mask_cls_sep',
                                     '-cpu'])`
`server = BertServer(args)`
`print("bert server start....")`
`server.start()`

`bc = BertClient(port=5555,port_out=5556)`
`result = bc.encode(['First do it', 'then do it right', 'then do it better'])`

`print(result)`

`BertServer.shutdown(port=5555)`
`print("bert server stop....")`

Error:
`Traceback (most recent call last):`
  `File "<input>", line 1, in <module>`
  `File "/home/cn-pvgtoddlu/.pycharm_helpers/pydev/_pydev_bundle/pydev_umd.py", line 197, in runfile
    pydev_imports.execfile(filename, global_vars, local_vars)  # execute the script`
  `File "/home/cn-pvgtoddlu/.pycharm_helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile`
    `exec(compile(contents+"\n", file, 'exec'), glob, loc)`
  `File "/home/cn-pvgtoddlu/tmp/pycharm_project_815/AI_Factory_Core/component/word_embedding/extract_embedding.py", line 21, in <module>`
    `BertServer.shutdown(port=5555)`
`TypeError: shutdown() got an unexpected keyword argument 'port'`

**args** of staticmethod shutdown are not from parser i think.

shutdown function:
`@staticmethod`
    `def shutdown(args):`
        `with zmq.Context() as ctx:`
            `ctx.setsockopt(zmq.LINGER, args.timeout)`
            `with ctx.socket(zmq.PUSH) as frontend:`
                `try:`
                    `frontend.connect('tcp://%s:%d' % (args.ip, args.port))`

lu161513:
I think it should be :

shut_args = get_shutdown_parser().parse_args(['-ip','localhost','-port','5555','-timeout','5000'])
BertServer.shutdown(shut_args)

lu161513:
我可以修正这个，可以merge进去吗~

[...]

PR #463, opened 2019-10-17
fix readme about how to use BertServer.shutdown with static method

lu161513:
fix issue #394

BertServer.shutdown() should use with args.
so fix readme


hanxiao:
👍 thanks a lot for your contribution

hanxiao:
fix #394

PR #486, opened 2019-11-25
Added server command line option for position embeddings, no_position…

shushanxingzhe:
…_embeddings can disable bert use_position_embeddings option, it is useful when different word order have the same meaning in short text area such as search query

### DIFF
--- a/.github/workflows/pythonpublish.yml
+++ b/.github/workflows/pythonpublish.yml
@@ -1,0 +1,26 @@
+name: bert-service-server
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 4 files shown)


hanxiao:
wow, thanks for setting up Github action for that! really appreciated!

but I do need some time to check and make sure it works as expected, let me first review the other two files.

hanxiao:
the other two files look fine. thanks and great job!

regarding the release pipeline, I actually use `./release.sh` for releasing, this includes bumping version, some chore jobs (re-organizing docs), etc. 

right now I will merge it as this Github action file wont affect the release procedure, and people may improve on it in the future.

All in all, thanks for your contribution. ❤️

hanxiao:
fyi, your contrib is now available at 1.9.9 and has been released on PyPI

PR #496, opened 2019-12-22
updated in/out ports to match the default values

DmitryKey:
This PR fixes an issue with connecting client to the server.


hanxiao:
thanks for your contribution ❤️

DmitryKey:
Thanks for accepting my PR and thank you for the great set of tools for BERT.

PR #506, opened 2020-01-21
OpenAPI contract and SwaggerUI support for bert-as-service

drndos:
Hello,
I have integrated [SwaggerUI](https://swagger.io/tools/swagger-ui/) into Flask http server and defined a [OpenAPI contract definition](https://swagger.io/docs/specification/about/) to support testing the API from browser and [generating clients for various languages.](https://github.com/OpenAPITools/openapi-generator) 

Open issues:
- Readme should be updated.
- Maybe it could be optional?
- Tokenization result is not present (it wasn't in http implementation either so it is not in contract)
- I have entered my info into contract, feel free to change it to something more appropriate

I am opened to feedback.
Thank you,
Filip

### DIFF
--- a/server/bert_serving/server/bertApi.openapi.yaml
+++ b/server/bert_serving/server/bertApi.openapi.yaml
@@ -1,0 +1,247 @@
+openapi: "3.0.2"
+info:
+  version: 1.0.0
+  title: Bert as a Service API
+  description: API to encode sentence or tokens to vectors
+  termsOfService: http://swagger.io/terms/
+  contact:
+    name: Filip Bednárik
+    email: info@ardevop.sk
+    url: https://ardevop.sk
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+servers:
+  - url: http://localhost:8125/
+paths:
+  /encode:
+    post:
+      description: |
+        Encode a list of strings to a list of vectors
+      operationId: app.encode_query
+      requestBody:
+        description: EncodeRequest to apply
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/EncodeRequest'
+      responses:
+        '200':
+          description: Encode response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/EncodeResponse'
+        default:
+          description: unexpected error
+  /status/server:
+    get:
+      description: Returns server status
+      operationId: get_server_status
+      responses:
+        '200':
+          description: Server status response
+          content:
+            application/json:
+              schema:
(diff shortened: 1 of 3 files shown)


hanxiao:
❤️ thanks a lot for your contribution! Really appreciated!

drndos:
Thank you for merging. Today after extended testing, I have found some issues, that I will be resolving soon (I will take a little bit more time testing this time). Namely polymorphism for tokenized and untokenized input is not present in current schema, also there is typo in property name for result(s).

hanxiao:
❤️  no problemo, feel free to continue on this thread and I will publish this change in the next release.

drndos:
Hello, I have come to a problem. Let me explain so we can decide what to do:

Endpoint /encode is capable of encoding either tokens or sentences. So two data structures can be part of the request, either {id:0, is_tokenized:true,texts:[['string']]} or {id:0, is_tokenized:false, texts['string']}

This is ok, but the issue lies in the OpenAPI generator and Swagger UI. There are two open issues:
- Swagger UI cannot generate examples when you use "oneOf" (which is used in polymorphism)
- OpenAPI Generator cannot generate the object structure properly when in request schema there is only "oneOf"

If I would model the request correctly, so that specification is foolproof then these two issues (that have open PRs) appear.
```
schema:
              oneOf:
                - $ref: '#/components/schemas/EncodeRequestTokenized'
                - $ref: '#/components/schemas/EncodeRequestNotTokenized'
              discriminator:
                propertyName: is_tokenized
                mapping:
                  true: '#/components/schemas/EncodeRequestTokenized'
                  false: '#/components/schemas/EncodeRequestNotTokenized'
              examples:
                EncodeRequestTokenized:
                  value:
                    id: 0
                    is_tokenized: true
                    texts: [['string']]
                EncodeRequestNotTokenized:
                  value:
                    id: 0
                    is_tokenized: false
                    texts: ['string']
```

If I would model it semi-correctly the schema would validate {id:0,is_tokenized:true} as correct.
```
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/EncodeRequest'
...
    EncodeRequest:
...
        is_tokenized:
          type: boolean
          default: false
      discriminator:
        propertyName: is_tokenized
        mapping:
          true: '#/components/schemas/EncodeRequestTokenized'
          false: '#/components/schemas/EncodeRequestNotTokenized'
    EncodeRequestTokenized:
      description: Also need to set the property is_tokenized to true
      allOf:
        - $ref: '#/components/schemas/EncodeRequest'
```
Then there is option to change the API itself to /encodeTokens and /encodeSentences - this seems like best solution, but is breaking change.

What do you think?
Thank you

Issue #567, opened 2020-06-24
Docker build failed creating '/app'


YoongiKim:
**Prerequisites**

* [x] Are you running the latest `bert-as-service`?
* [x] Did you follow [the installation](https://github.com/hanxiao/bert-as-service#install) and [the usage](https://github.com/hanxiao/bert-as-service#usage) instructions in `README.md`?
* [x] Did you check the [FAQ list in `README.md`](https://github.com/hanxiao/bert-as-service#speech_balloon-faq)?
* [x] Did you perform [a cursory search on existing issues](https://github.com/hanxiao/bert-as-service/issues)?

**System information**
- Using docker image: tensorflow/tensorflow:1.12.0-gpu-py3

---

### Description

> **How to reproduce**

```bash
git clone https://github.com/hanxiao/bert-as-service.git
cd bert-as-service
docker build -t bert-as-service -f ./docker/Dockerfile .
```

> **Output**

```bash
~/Projects/bert-as-service$ docker build -t bert-as-service -f ./docker/Dockerfile .
Sending build context to Docker daemon  13.45MB
Step 1/7 : FROM tensorflow/tensorflow:1.12.0-gpu-py3
 ---> 413b9533f92a
Step 2/7 : RUN pip install bert-serving-server[http]
 ---> Using cache
 ---> 84941411ab9c
Step 3/7 : COPY ./docker/entrypoint.sh /app
 ---> Using cache
 ---> e2fbe8e0b7ca
Step 4/7 : WORKDIR /app
Cannot mkdir: /app is not a directory
```
Error message is **Cannot mkdir: /app is not a directory**

> Original docker/Dockerfile
```
FROM tensorflow/tensorflow:1.12.0-gpu-py3
RUN pip install bert-serving-server[http]
COPY ./docker/entrypoint.sh /app
WORKDIR /app
ENTRYPOINT ["/app/entrypoint.sh"]
CMD []
HEALTHCHECK --timeout=5s CMD curl -f http://localhost:8125/status/server || exit 1
```

If I add command ```RUN ls -al /``` before ```WORKDIR /app```
```bash
drwxr-xr-x   1 root root         6 Jun 24 08:16 .
drwxr-xr-x   1 root root         6 Jun 24 08:16 ..
-rwxr-xr-x   1 root root         0 Jun 24 08:16 .dockerenv
**-rwxrwxr-x   1 root root        78 Jun 24 08:01 app**
[...]

PR #568, opened 2020-06-24
Fix docker build error. #567

YoongiKim:
Fix '/app' to '/app/'


mdoulaty:
this is a duplicate of https://github.com/hanxiao/bert-as-service/pull/560

PR #615, opened 2021-01-01
Apply tensorflow 2.0's v1 compatibility layer so the code works with TF2.

mmohaveri:
The current master does not work with Tensorflow 2.0 because it's using a set of deprecated APIs.
This PR will replace those deprecated APIs with the ones from "tf.compat.v1".

### DIFF
--- a/server/bert_serving/server/__init__.py
+++ b/server/bert_serving/server/__init__.py
@@ -491,8 +491,8 @@ class BertWorker(Process):
         from tensorflow.python.estimator.model_fn import EstimatorSpec
 
         def model_fn(features, labels, mode, params):
-            with tf.gfile.GFile(self.graph_path, 'rb') as f:
-                graph_def = tf.GraphDef()
+            with tf.io.gfile.GFile(self.graph_path, 'rb') as f:
+                graph_def = tf.compat.v1.GraphDef()
                 graph_def.ParseFromString(f.read())
 
             input_names = ['input_ids', 'input_mask', 'input_type_ids']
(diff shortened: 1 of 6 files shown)


AndreaSottana:
Hi @mmohaveri 
I have tried using your code. However, tf is not able to build the graph correctly, at least for me.
I have downloaded the BER model, I then open a terminal and run the following
`bert-serving-start -model_dir=/d/Users/[...]/uncased_L-12_H-768_A-12 -cpu`

But I then get the following error
```
Traceback (most recent call last):
  File "D:\Users\[...]\venv38\Scripts\bert-serving-start-script.py", line 33, in <module>
    sys.exit(load_entry_point('bert-serving-server', 'console_scripts', 'bert-serving-start')())
  File "d:\users\[...]\bert-as-service\server\bert_serving\server\cli\__init__.py", line 8, in main
    with BertServer(get_run_args()) as server:
  File "d:\users\[...]\bert-as-service\server\bert_serving\server\__init__.py", line 75, in __init__
    self.graph_path, self.bert_config = pool.apply(optimize_graph, (self.args,))
TypeError: cannot unpack non-iterable NoneType object
```

even though the Bert model is there. Any idea what might be going wrong?

Akila-Ayanthi:
@AndreaSottana were you able to get it to work?

Akila-Ayanthi:
@mmohaveri How do u run BAS from the source code?

AndreaSottana:
@Akila-Ayanthi No I didn't, I suspect the original code from hanxiao:master (last commit August 2020) is not 
regularly maintained and something might have broken due to version issues.
I am using this very good alternative though, to integrate BERT with ElasticSearch https://github.com/UKPLab/sentence-transformers/blob/master/examples/applications/semantic-search/semantic_search_quora_elasticsearch.py

Akila-Ayanthi:
@AndreaSottana  Thank you.

PR #663, opened 2022-03-24
perf: use dataloader to avoid GPU waiting for workers

alaeddine-13:
Encoding and preprocessing are executed sequentially and the GPU waits until all docs are ready to be encoded.
@samsja and I wanted to avoid this by using the DataLoader from pytorch so that workers contribute in processing batches to be encoded.

`python profile_executor_text.py`

On master:
```text
cpu preprocessing 0.5129575690007186
gpu embedding 2.19830906199968
first time 2.9628389139998035
cpu preprocessing 0.5007643369999641
gpu embedding 2.184567760000391
second time 2.9293103939999128
```

feat:
```text
# feat dataloader
gpu waiting for workers 0.07661872600056086
gpu encoding 0.22060704299929057
gpu waiting for workers 0.0021561820003626053
gpu encoding 0.22138791300039884
gpu waiting for workers 0.0010702899999159854
gpu encoding 0.22121322800012422
gpu waiting for workers 9.291199967265129e-05
gpu encoding 0.22160035900014918
gpu waiting for workers 8.87120004335884e-05
gpu encoding 0.22376149100000475
gpu waiting for workers 8.738199994695606e-05
gpu encoding 0.22122334699997737
gpu waiting for workers 0.000711570000021311
gpu encoding 0.221622846000173
gpu waiting for workers 0.0011233209997953963
gpu encoding 0.2221713130002172
gpu waiting for workers 8.407299992541084e-05
gpu encoding 0.22373632800008636
gpu waiting for workers 9.173199941869825e-05
gpu encoding 0.2231302689997392
second time 2.5170462010000847
```

Currently, spawning processes for the first time takes around 1 second that's why we only care about the second execution.

### DIFF
--- a/server/clip_server/executors/clip_torch_dataloader.py
+++ b/server/clip_server/executors/clip_torch_dataloader.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING, Optional, List
 import torch
 from PIL import Image
 from jina import Executor, requests
+from torch.utils.data import Dataset, DataLoader
 
 from clip_server.model import clip
 
@@ -13,13 +14,31 @@ if TYPE_CHECKING:
     from docarray import DocumentArray, Document
 
 
+class DocumentDataset(Dataset):
+    def __init__(self, device):
+        self.device = device
+        self.da = None
+
+    def set_da(self, da: 'DocumentArray'):
+        self.da = da
+
+    def __len__(self):
+        return len(self.da)
+
+    def __getitem__(self, idx):
+        text = self.da[idx].text
+        tensor = clip.tokenize(text)
+        return tensor[0]
+
+
 class CLIPEncoder(Executor):
     def __init__(
-        self,
-        name: str = 'ViT-B/32',
-        device: Optional[str] = None,
-        jit: bool = False,
-        **kwargs
+            self,
+            name: str = 'ViT-B/32',
+            device: Optional[str] = None,
+            jit: bool = False,
+            batch_size: int = 1000,
+            **kwargs
     ):
         super().__init__(**kwargs)
         if not device:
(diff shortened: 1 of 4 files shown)


hanxiao:
great job on inspecting this.

@samsja btw if core can somehow enable the profiling for any executor that would be awesome! basically auto injecting the similar timing code that @alaeddine-13 wrote into Executor

bwanglzu:
when using `DataLoader` please try [num_workers](https://discuss.pytorch.org/t/guidelines-for-assigning-num-workers-to-dataloader/813/5) and [pin_memory=True](https://developer.nvidia.com/blog/how-optimize-data-transfers-cuda-cc/)

server/clip_server/executors/clip_torch_dataloader.py
```diff
@@ -0,0 +1,92 @@
...
+        super().__init__(**kwargs)
+        if not device:
+            self._device = 'cuda' if torch.cuda.is_available() else 'cpu'
+        else:
+            self._device = device
+        self._model, self._preprocess = clip.load(name, device=self._device, jit=jit)
+        self.batch_size = batch_size
+
+        self.dataset = DocumentDataset(self._device)
+        self.dataloader = DataLoader(self.dataset, batch_size=self.batch_size, num_workers=48, persistent_workers=True)
```
numb3r3, line 52: Can we confirm the performance diff doesn't come from `num_workers=48`?

numb3r3: it does not mean more workers will definitely lead to faster processing speed. Generally, it should not exceed the number of CPU cores we have.

samsja: we have 64 cpu core in the berlin machine. In our case we have to prepare 10 batch so we could just stick with 10 workers

samsja:
> great job on inspecting this.
> 
> @samsja btw if core can somehow enable the profiling for any executor that would be awesome! basically auto injecting the similar timing code that @alaeddine-13 wrote into Executor

server/clip_server/executors/clip_torch_dataloader.py
numb3r3: If the improvement mainly comes from `num_workers=48`, then we can utilize `da.apply(..., num_workers=48)` implemented in `docarray` instead.

PR #665, opened 2022-03-28
docs(tracking): add scarf tracking

Roshanjossey:


numb3r3:
For this PR, could you provide more context? I just found it is the first time we use  `Scarf` in our projects. It is a toolchain for maintainers. Do we need to set up other components, e.g., some service, provider?

alexcg1 (approved):
LGTM :+1: 

Note to other reviewers: scarf tracking gives us more granularity (e.g. company name) into who is reading our docs

Assigning whoever is head of this workstream to merge, since it's not my codebase

PR #744, opened 2022-06-09
fix: jit execution optimization

numb3r3:


numb3r3:
I will close this PR, since no improvement can be archived with this changes.

PR #748, opened 2022-06-10
feat: add traversal paths

numb3r3:


server/clip_server/executors/clip_onnx.py
```diff
@@ -59,7 +61,7 @@ def __init__(
             and hasattr(self.runtime_args, 'replicas')
         ):
             replicas = getattr(self.runtime_args, 'replicas', 1)
-            num_threads = max(1, torch.get_num_threads() // replicas)
+            num_threads = max(1, torch.get_num_threads() * 2 // replicas)
```
ZiniuYu, line 64: why *2

numb3r3: 50% CPU utilization with * 1

ZiniuYu (commented):
Need to add some test cases

client/clip_client/client.py
```diff
@@ -184,12 +187,18 @@ def _iter_doc(self, content) -> Generator['Document', None, None]:
...
+
+        payload = dict(
             on='/',
             inputs=self._iter_doc(content),
-            request_size=kwargs.get('batch_size', 8),
+            request_size=kwargs.get('batch_size', 32),
+            parameters=parameters,
             total_docs=len(content) if hasattr(content, '__len__') else None,
         )
+        return payload
```
ZiniuYu, line 203: combine these lines

ZiniuYu (changes requested):

ZiniuYu (approved):
LGTM

hanxiao:
traversal path should not be handled in CAS, user should first traverse and then send via client

client/clip_client/client.py
```diff
@@ -158,17 +158,13 @@ def _iter_doc(self, content) -> Generator['Document', None, None]:
...
             elif isinstance(c, Document):
+                self._return_plain = False
                 if c.content_type in ('text', 'blob'):
-                    self._return_plain = False
                     yield c
                 elif not c.blob and c.uri:
                     c.load_uri_to_blob()
-                    self._return_plain = False
-                    yield c
-                elif c.tensor is not None:
```
ZiniuYu, line 168: should keep this line

numb3r3:
> traversal path should not be handled in CAS, user should first traverse and then send via client

The executors in the CAS will then be uploaded to the hub. So I think the `traversal_paths` support is necessary.  And of course, if the user uses CAS client, the recommended way is to traverse first (the `traversal_paths` will not be mentioned in CAS doc).

PR #751, opened 2022-06-13
docs: add disk usage / memory usage benchmark table

ZiniuYu:


docs/user-guides/server.md
```diff
@@ -62,17 +62,17 @@ The procedure and UI of ONNX and TensorRT runtime would look the same as Pytorch
...
-| RN50 | ✅ |✅ | ✅| 1024 | 
-| RN101 | ✅ |✅ | ✅| 512 | 
-| RN50x4 | ✅ |✅ | ✅| 640 |
-| RN50x16 | ✅ |✅ | ❌| 768 |
-| RN50x64 | ✅ |✅ | ❌| 1024 |
-| ViT-B/32  | ✅ |✅ | ✅| 512 |
-| ViT-B/16 | ✅ |✅ | ✅| 512 |
-| ViT-L/14 | ✅ |✅ | ✅| 768 |
-| ViT-L/14-336px | ✅ |✅ | ❌| 768 |
+| Model          | PyTorch | ONNX | TensorRT | Output dimension | Disk Usage | Peak Memory Usage |
```
numb3r3, line 65: `Disk Usage` please also explain the memory unit. For `Peak Memory Usage`, please also provide what's the context, e.g., the GPU device info, the batch size

numb3r3 (changes requested):

docs/user-guides/server.md
ZiniuYu: Yes right now it is just a placeholder, am adding more details

docs/user-guides/server.md
```diff
@@ -60,19 +60,19 @@ The procedure and UI of ONNX and TensorRT runtime would look the same as Pytorch
...
+|----------------|---------|------|----------|------------------|-----------------|------------------------|
+| RN50           | ✅       | ✅    | ✅        | 1024             | 256             | 4.25                   |
+| RN101          | ✅       | ✅    | ✅        | 512              | 292             | 4.28                   |
+| RN50x4         | ✅       | ✅    | ✅        | 640              | 422             | 6.92                   |
+| RN50x16        | ✅       | ✅    | ❌        | 768              | 661             | 13.01                  |
+| RN50x64        | ✅       | ✅    | ❌        | 1024             | 1382            | 20.26                  |
+| ViT-B/32       | ✅       | ✅    | ✅        | 512              | 351             | 2.19                   |
+| ViT-B/16       | ✅       | ✅    | ✅        | 512              | 354             | 3.90                   |
+| ViT-L/14       | ✅       | ✅    | ✅        | 768              | 933             | 5.38                   |
+| ViT-L/14-336px | ✅       | ✅    | ❌        | 768              | 934             | 11.36                  |
```
hanxiao, line 75: what about VRAM? thats very important.

docs/user-guides/server.md
```diff
@@ -60,19 +60,19 @@ The procedure and UI of ONNX and TensorRT runtime would look the same as Pytorch
...
+| RN50           | ✅       | ✅    | ✅        | 1024             | 256             | 2.97                | 4.25                |
+| RN101          | ✅       | ✅    | ✅        | 512              | 292             | 3.05                | 4.28                |
+| RN50x4         | ✅       | ✅    | ✅        | 640              | 422             | 3.18                | 6.92                |
+| RN50x16        | ✅       | ✅    | ❌        | 768              | 661             | 3.55                | 13.01               |
+| RN50x64        | ✅       | ✅    | ❌        | 1024             | 1382            | 3.95                | 20.26               |
+| ViT-B/32       | ✅       | ✅    | ✅        | 512              | 351             | 3.18                | 2.19                |
+| ViT-B/16       | ✅       | ✅    | ✅        | 512              | 354             | 3.17                | 3.90                |
+| ViT-L/14       | ✅       | ✅    | ✅        | 768              | 933             | 3.64                | 5.38                |
+| ViT-L/14-336px | ✅       | ✅    | ❌        | 768              | 934             | 3.67                | 11.36               |
```
hanxiao, line 75: on RAM and vRAM, either the number is wrong or the column names are wrong. Please check again

docs/user-guides/server.md
```diff
@@ -60,19 +60,19 @@ The procedure and UI of ONNX and TensorRT runtime would look the same as Pytorch
 
 ## Model support
 
-Open AI has released 9 models so far. `ViT-B/32` is used as default model in all runtimes. Due to the limitation of some runtime, not every runtime supports all nine models. Please also note that different model give different size of output dimensions. This will affect your downstream applications. For example, switching the model from one to another make your embedding incomparable, which breaks the downstream applications. Here is a list of supported models of each runtime and its corresponding size:
+Open AI has released 9 models so far. `ViT-B/32` is used as default model in all runtimes. Due to the limitation of some runtime, not every runtime supports all nine models. Please also note that different model give different size of output dimensions. This will affect your downstream applications. For example, switching the model from one to another make your embedding incomparable, which breaks the downstream applications. Below is a list of supported models of each runtime and its corresponding size. We also include a sample disk usage and peak memory usage when running with a single Nvidia TITAN RTX GPU with 24 GB memory and with a batch size of 256.
```
hanxiao, line 63:
```suggestion
Open AI has released 9 models so far. `ViT-B/32` is used as default model in all runtimes. Due to the limitation of some runtime, not every runtime supports all nine models. Please also note that different model give different size of output dimensions. This will affect your downstream applications. For example, switching the model from one to another make your embedding incomparable, which breaks the downstream applications. Below is a list of supported models of each runtime and its corresponding size. We include the disk usage (in delta) and the peak memory usage (in delta) when running on a single Nvidia TITAN RTX GPU (24GB VRAM) when `batch_size=256`.
```

docs/user-guides/server.md
ZiniuYu: I quickly reran the test and noticed that the RAM usage is under-calculated by ~20%. The table is updated

docs/user-guides/server.md
```diff
@@ -60,19 +60,19 @@ The procedure and UI of ONNX and TensorRT runtime would look the same as Pytorch
...
+| RN50           | ✅       | ✅    | ✅        | 1024             | 256             | 3.44                | 4.25                 |
+| RN101          | ✅       | ✅    | ✅        | 512              | 292             | 3.51                | 4.28                 |
+| RN50x4         | ✅       | ✅    | ✅        | 640              | 422             | 4.14                | 6.92                 |
+| RN50x16        | ✅       | ✅    | ❌        | 768              | 661             | 5.07                | 13.01                |
+| RN50x64        | ✅       | ✅    | ❌        | 1024             | 1382            | 5.81                | 20.26                |
+| ViT-B/32       | ✅       | ✅    | ✅        | 512              | 351             | 3.66                | 2.19                 |
+| ViT-B/16       | ✅       | ✅    | ✅        | 512              | 354             | 3.58                | 3.90                 |
+| ViT-L/14       | ✅       | ✅    | ✅        | 768              | 933             | 4.11                | 5.38                 |
+| ViT-L/14-336px | ✅       | ✅    | ❌        | 768              | 934             | 5.07                | 11.36                |
```
hanxiao, line 75: image

This number is impossible. In general, I believe the whole column is wrong.
- We run 2 `ViT-L/14-336px` on Berlin GPU for the last 2 months and together they take 5GB VRAM. 
- see also similar usage on Google Colab https://jina-ai.slack.com/archives/CTH1CMP8V/p1655126173106729
image

ZiniuYu: In fact, the VRAM data is not wrong. I think the reason I got a large VRAM value is that I was using a mini-batch of size 256 in clip-server. If I stick to the default value, the VRAM indeed drops to ~4GB.

ZiniuYu: I think I should clarify on that in the docs

ZiniuYu: I see. Using a different batch size indeed introduces some confusion. Will update the table

numb3r3:
@ZiniuYu please fix the commit message issue. 
```
Error: You have commit messages with errors
⧗   input: docs: RAM usage
✖   subject must not be sentence-case, start-case, pascal-case, upper-case [subject-case]
✖   found 1 problems, 0 warnings
```

docs/user-guides/server.md
```diff
@@ -60,19 +60,19 @@ The procedure and UI of ONNX and TensorRT runtime would look the same as Pytorch
...
+Open AI has released 9 models so far. `ViT-B/32` is used as default model in all runtimes. Due to the limitation of some runtime, not every runtime supports all nine models. Please also note that different model give different size of output dimensions. This will affect your downstream applications. For example, switching the model from one to another make your embedding incomparable, which breaks the downstream applications. Below is a list of supported models of each runtime and its corresponding size. We include the disk usage (in delta) and the peak RAM and VRAM usage (in delta) when running on a single Nvidia TITAN RTX GPU (24GB VRAM) using a default `batch_size=32` in server and a default `batch_size=8` in client.
```
hanxiao, line 63: i believe on the server it is not called `batch_size` but `mini_batch_size`?

hanxiao (approved):
looks like a test failing

PR #752, opened 2022-06-13
fix: fix client concurrent issue

ZiniuYu:
In the original client implementation, it uses a global result variable to store all results from the server. This may break with the async client in that it can not distinguish the source of input and thus mix all concurrent requests and return the wrong results.

This PR fixes the above issue.


client/clip_client/client.py
```diff
@@ -109,19 +111,21 @@ def encode(self, content, **kwargs):
...
         )
+        results = DocumentArray()
         with self._pbar:
             self._client.post(
-                **self._get_post_payload(content, kwargs), on_done=self._gather_result
+                **self._get_post_payload(content, kwargs),
+                on_done=partial(self._gather_result, results=results),
             )
-        return self._unboxed_result
+        return self._unboxed_result(results, self._return_plain)
```
numb3r3, line 120: `self._return_plain` suffer from the same issue as `self._results`

numb3r3 (changes requested):

client/clip_client/client.py
```diff
@@ -130,15 +134,15 @@ def _gather_result(self, r):
             ),
         )
 
-    @property
-    def _unboxed_result(self):
-        if self._results.embeddings is None:
+    @staticmethod
+    def _unboxed_result(results, return_plain):
```
numb3r3, line 138:
```suggestion
    def _unboxed_result(results, return_plain: bool):
```

client/clip_client/client.py
ZiniuYu: Good point

client/clip_client/client.py
```diff
@@ -130,40 +128,38 @@ def _gather_result(self, r):
...
         if hasattr(self, '_pbar'):
             self._pbar.start_task(self._s_task)
 
         for c in content:
             if isinstance(c, str):
-                self._return_plain = True
                 _mime = mimetypes.guess_type(c)[0]
                 if _mime and _mime.startswith('image'):
-                    yield Document(uri=c).load_uri_to_blob()
+                    yield Document(tags={'_plain': True}, uri=c).load_uri_to_blob()
```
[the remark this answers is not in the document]
numb3r3: And make sure the unittest can cover this case.

numb3r3 (changes requested):

numb3r3 (approved):
LGTM

PR #753, opened 2022-06-14
fix: add integerate test for client

numb3r3:


tests/test_client.py
```diff
@@ -0,0 +1,57 @@
...
+        def generate_docs(tag):
+            return DocumentArray(
+                [Document(id=f'{tag}_{i}', text='hello') for i in range(20)]
+            )
+
+        with ThreadPool(5) as p:
+            results = p.map(run_post, [generate_docs(f't{k}') for k in range(5)])
+
+        for r in results:
+            assert len(set([d.id[:2] for d in r])) == 1
```
ZiniuYu, line 57: What about using asyncio.create_task like what is used in test_asyncio?

numb3r3: `asyncio` looks more fancy.  But I won't change my code if there is no problem with threading.

ZiniuYu (approved):

PR #763, opened 2022-07-11
feat: support multilingual clip

shan-mx:
This PR allows the user to use an clip executor that supports multilingual sentences.

### DIFF
--- a/server/clip_server/executors/mclip_hg.py
+++ b/server/clip_server/executors/mclip_hg.py
@@ -57,6 +57,12 @@ class CLIPEncoder(Executor):
     ):
         """
         :param pretrained_model_name_or_path: Can be either:
+            - A string, the model id of a pretrained M-CLIP model hosted
+                inside a model repo on huggingface.co, e.g.,
+                'M-CLIP/XLM-Roberta-Large-Vit-B-32'
+            - A path to a directory containing model weights saved, e.g.,
+                ./my_model_directory/
+        :param clip_model_name_or_path: Can be either:
             - A string, the model id of a pretrained CLIP model hosted
                 inside a model repo on huggingface.co, e.g.,
                 'openai/clip-vit-base-patch32'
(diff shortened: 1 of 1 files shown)


server/clip_server/executors/mclip_hg.py
```diff
@@ -0,0 +1,276 @@
...
+        att = txt_tok['attention_mask']
+        embs = (embs * att.unsqueeze(2)).sum(dim=1) / att.sum(dim=1)[:, None]
+        return self.LinearTransformation(embs)
+
+
+class CLIPEncoder(Executor):
+    def __init__(
+        self,
+        pretrained_model_name_or_path: str = 'M-CLIP/XLM-Roberta-Large-Vit-B-32',
+        clip_model_name_or_path: str = 'openai/clip-vit-base-patch32',
```
numb3r3, line 45: For the `M-CLIP` style models, is there a list of the variate models? I guess we need maintain a model list to tell the user how to chose the right model regarding to their use case.

numb3r3: What's more, can we kill the `pretrain_model_name_or_path`, and share the `clip_model_name` parameter to specify which model to load?

server/clip_server/executors/mclip_hg.py
```diff
@@ -0,0 +1,276 @@
...
+    split_img_txt_da,
+    set_rank,
+)
+from jina import Executor, requests, DocumentArray
+
+
+class MCLIPConfig(PretrainedConfig):
+    model_type = "M-CLIP"
+
+    def __init__(self, modelBase='xlm-roberta-large', transformerDimSize=1024, imageDimSize=768, **kwargs):
```
numb3r3, line 18: Don't use camel case for the variables/parameters, e.g., `transformerDimSize`. This name convention does not match the style in the current code base.

numb3r3 (changes requested):
Basically, this PR looks good. To merge this PR, there is some comments that need to be addressed. And what's more, please follow this guideline https://github.com/jina-ai/docarray/blob/main/CONTRIBUTING.md.

And please remember to install the pre-commit before commit:
```
pre-commit install
```

PR #765, opened 2022-07-12
feat: allow credential in client

ZiniuYu:
This pr allow users to pass credential for HTTP and GRPC, and use it for authentication when getting request from server

### DIFF
--- a/client/clip_client/client.py
+++ b/client/clip_client/client.py
@@ -192,7 +192,7 @@ class Client:
             total_docs=len(content) if hasattr(content, '__len__') else None,
         )
         if self._scheme == 'grpc' and self._authorization:
-            payload.update(metadata=('Authorization', self._authorization))
+            payload.update(metadata=('authorization', self._authorization))
         elif self._scheme == 'http' and self._authorization:
             payload.update(headers={'Authorization': self._authorization})
         return payload
@@ -373,7 +373,7 @@ class Client:
             total_docs=len(content) if hasattr(content, '__len__') else None,
         )
         if self._scheme == 'grpc' and self._authorization:
-            payload.update(metadata=('Authorization', self._authorization))
+            payload.update(metadata=('authorization', self._authorization))
         elif self._scheme == 'http' and self._authorization:
             payload.update(headers={'Authorization': self._authorization})
         return payload
--- a/client/clip_client/client.py
+++ b/client/clip_client/client.py
@@ -15,6 +15,7 @@ from typing import (
 from urllib.parse import urlparse
 from functools import partial
 from docarray import DocumentArray
+from helper import _grpc_credential_wrapper, _http_credential_wrapper
 
 if TYPE_CHECKING:
     import numpy as np
@@ -22,7 +23,7 @@ if TYPE_CHECKING:
 
 
 class Client:
-    def __init__(self, server: str):
+    def __init__(self, server: str, credential: str = None, **kwargs):
         """Create a Clip client object that connects to the Clip server.
         Server scheme is in the format of `scheme://netloc:port`, where
             - scheme: one of grpc, websocket, http, grpcs, websockets, https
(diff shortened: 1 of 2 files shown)


client/clip_client/client.py
```diff
@@ -30,26 +30,23 @@ def __init__(self, server: str):
...
-            r = urlparse(server)
-            _port = r.port
-            _scheme = r.scheme
-            if not _scheme:
-                raise
-        except:
+        r = urlparse(server)
+        _port = r.port
+        self.scheme = r.scheme
+        if not self.scheme:
```
numb3r3, line 36: how to catch the exception during `urlparse(server)`?

client/clip_client/client.py
```diff
@@ -30,26 +30,23 @@ def __init__(self, server: str):
...
 
-        if _scheme in ('grpcs', 'https', 'wss'):
-            _scheme = _scheme[:-1]
+        if self.scheme in ('grpcs', 'https', 'wss'):
+            self.scheme = self.scheme[:-1]
             _tls = True
 
-        if _scheme == 'ws':
-            _scheme = 'websocket'  # temp fix for the core
+        if self.scheme == 'ws':
```
numb3r3, line 45: use `self._schema`

numb3r3 (changes requested):

client/clip_client/client.py
ZiniuYu: it would fail either way since we cannot create a connection from a broken link

client/clip_client/client.py
```diff
@@ -181,18 +181,17 @@ def _iter_doc(self, content) -> Generator['Document', None, None]:
...
-            else None,
-            headers=_http_credential_wrapper(self.credential)
-            if self._scheme == 'http'
-            else None,
         )
+        if self._scheme == 'grpc':
+            payload.update(metadata=get_authentication(self._scheme, self.credential))
+        elif self._scheme == 'http':
+            payload.update(headers=get_authentication(self._scheme, self.credential))
+        return payload
```
numb3r3, line 194: If WebSocket does not accept credential, then let's raise an exception to tell the user.

client/clip_client/helper.py
```diff
@@ -54,9 +54,18 @@ def is_latest_version(package: str = None, github_repo: str = None) -> None:
     threading.Thread(target=_version_check, args=(package, github_repo)).start()
 
 
-def _http_credential_wrapper(credential):
-    return {'Authorization': f'{credential}'}
-
-
-def _grpc_credential_wrapper(credential):
-    return ('Authorization', f'{credential}')
+def get_authentication(scheme, credentials):
```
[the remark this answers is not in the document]
numb3r3: keep consistency of the parameter name, `crendentials` vs `crendential`

client/clip_client/client.py
```diff
@@ -361,20 +360,19 @@ def _iter_rank_docs(
...
             total_docs=len(content) if hasattr(content, '__len__') else None,
-            metadata=_grpc_credential_wrapper(self.credential)
-            if self._scheme == 'grpc'
-            else None,
-            headers=_http_credential_wrapper(self.credential)
-            if self._scheme == 'http'
-            else None,
         )
+        if self._scheme == 'grpc':
+            payload.update(metadata=get_authentication(self._scheme, self.credential))
```
numb3r3: how to process with the case of  `get_authentication()` returns None?

client/clip_client/client.py
```diff
@@ -33,23 +34,22 @@ def __init__(self, server: str):
...
             r = urlparse(server)
             _port = r.port
-            _scheme = r.scheme
-            if not _scheme:
-                raise
+            self._scheme = r.scheme
         except:
             raise ValueError(f'{server} is not a valid scheme')
 
+        self.credential = credential
```
numb3r3, line 41: should be `self._crendential = credential or {}`

client/clip_client/client.py
ZiniuYu: None value wont affect

client/clip_client/helper.py
```diff
@@ -52,3 +52,20 @@ def is_latest_version(package: str = None, github_repo: str = None) -> None:
...
+            else None
+        )
+    elif scheme == 'http':
+        return (
+            {'Authorization': credential['Authorization']}
+            if 'Authorization' in credential
+            else None
+        )
+    else:
+        return None
```
numb3r3, line 71:
```suggestion
def get_authorization(scheme, credential):
    auth = ()
       
    if crendential and 'Authorization' in crendential:
        auth += (('Authorization': credential['Authorization']),)
    
    if schema == 'http':
        auth = dict(auth)
    return auth
```

numb3r3 (changes requested):

client/clip_client/helper.py
ZiniuYu: i think we don't need this and can make it one-liner

numb3r3 (changes requested):
This PR looks fine so far. But I will block this PR until **we have a solid integration test ready**.

numb3r3 (changes requested):

PR #767, opened 2022-07-14
feat: support multlingual clip

shan-mx:


server/clip_server/executors/clip_torch.py
```diff
@@ -5,25 +5,26 @@
...
     def __init__(
         self,
-        name: str = 'ViT-B/32',
+        name: str = "M-CLIP/XLM-Roberta-Large-Vit-B-32",
         device: Optional[str] = None,
         jit: bool = False,
         num_worker_preprocess: int = 4,
         minibatch_size: int = 32,
-        traversal_paths: str = '@r',
+        traversal_paths: str = "@r",
```
numb3r3, line 27: keep to use single quotation `'`

numb3r3 (changes requested):
I would like to reject this PR. We should have a good design first to proceed. The current implementation looks like a patch, rather than a good feature.

numb3r3:
We have a better idea to support this feature. I will close this PR now.

PR #770, opened 2022-07-15
feat: multi platform support v1

shan-mx:


numb3r3:
This PR is an experimental one.

PR #775, opened 2022-07-20
feat: allow model selection in client

ZiniuYu:
This pr allows the user to select the model in the client when multiple models are running.

Sample usage:

```
from clip_client import Client

c = Client(xxx)
c.encode(xxx, parameters={'model': 'RN50'})
c.rank(xxx, parameters={'model': 'RN50'})
```

### DIFF
--- a/client/clip_client/client.py
+++ b/client/clip_client/client.py
@@ -190,10 +190,9 @@ class Client:
 
     def _get_post_payload(self, content, kwargs):
         parameters = kwargs.get('parameters', {})
-        target_model = parameters.get('model', None)
-        endpoint = f'/encode{f"/{target_model}" if target_model else ""}'
+        model_name = parameters.get('model', '')
         payload = dict(
-            on=endpoint,
+            on=f'/encode/{model_name}',
             inputs=self._iter_doc(content),
             request_size=kwargs.get('batch_size', 8),
             total_docs=len(content) if hasattr(content, '__len__') else None,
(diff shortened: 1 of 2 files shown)


docs/user-guides/client.md
```diff
@@ -192,6 +192,65 @@ Use `.encode(..., show_progress=True)` to turn on the progress bar.
...
+  - name: ViT_B_32
+    replicas: 1
+    uses:
+      jtype: CLIPEncoder
+      with:
+        name: ViT-B/32
+      metas:
+        py_modules:
+          - clip_server.executors.clip_torch
+      requests: {'/encode/ViT-B/32': 'encode', '/': '_dry_run_func'}
```
numb3r3, line 220: Let's first hide this feature. This is a very advanced usage.

numb3r3 (changes requested):

client/clip_client/client.py
```diff
@@ -185,8 +189,11 @@ def _iter_doc(self, content) -> Generator['Document', None, None]:
                 )
 
     def _get_post_payload(self, content, kwargs):
+        parameters = kwargs.get('parameters', {})
+        target_model = parameters.get('model', None)
+        endpoint = f'/encode{f"/{target_model}" if target_model else ""}'
```
numb3r3, line 194:
```suggestion
        model_name = parameters.get('model', '')
        endpoint = f'/encode/{model_name}'
```

numb3r3 (changes requested):

client/clip_client/client.py
```diff
@@ -185,8 +189,11 @@ def _iter_doc(self, content) -> Generator['Document', None, None]:
...
         payload = dict(
-            on='/',
+            on=endpoint,
```
numb3r3, line 196:
```suggestion
            on=f'/encode/{model_name}',
```

numb3r3 (changes requested):

client/clip_client/client.py
ZiniuYu: smart

client/clip_client/client.py
ZiniuYu: are we introducing extra '/'?

numb3r3 (approved):

PR #777, opened 2022-07-22
fix: do not load non-local uri in client

ZiniuYu:


client/clip_client/client.py
```diff
@@ -115,10 +115,18 @@ def encode(self, content, **kwargs):
         )
         results = DocumentArray()
         with self._pbar:
-            self._client.post(
-                **self._get_post_payload(content, kwargs),
-                on_done=partial(self._gather_result, results=results),
-            )
+            try:
```
numb3r3, line 118: I just want to confirm will we have the same behavior during `aencode()` and `arank()`

numb3r3 (changes requested):

client/clip_client/client.py
ZiniuYu: no worry, working on it

PR #781, opened 2022-07-25
feat: support onnx backend for openclip

numb3r3:


server/clip_server/model/clip_model.py
```diff
@@ -5,7 +5,25 @@
...
+        super().__init__()
+        self._name = name
+
+    @staticmethod
+    def get_model_name(name: str):
+        return name
+
+    @property
+    def model_name(self):
+        return self.__class__.get_model_name(self._name)
```
ZiniuYu, line 19: can we simply return `xxx.name`?

numb3r3: To get the model name without initializing the instance (avoiding downloading and loading weights), we must use *staticmethod* . That's the reason I implement `get_model_name`

jemmyshin (approved):
LGTM!

PR #782, opened 2022-07-25
feat: update openclip loader

shan-mx:
to support independent download process and make precision adapted to device to solve VRAM issue

### DIFF
--- a/server/clip_server/model/mclip_model.py
+++ b/server/clip_server/model/mclip_model.py
@@ -74,4 +74,4 @@ class MultilingualCLIPModel(CLIPModel):
         )
 
     def encode_image(self, pixel_values: torch.Tensor, **kwargs):
-        return self._model.encode_image(pixel_values)
+        return self._model.encode_image(pixel_values, **kwargs)
--- a/server/clip_server/model/mclip_model.py
+++ b/server/clip_server/model/mclip_model.py
@@ -73,5 +73,5 @@ class MultilingualCLIPModel(CLIPModel):
             input_ids=input_ids, attention_mask=attention_mask, **kwargs
         )
 
-    def encode_image(self, pixel_values: torch.Tensor, **kwargs):
-        return self._model.encode_image(pixel_values, **kwargs)
+    def encode_image(self, pixel_values: torch.Tensor):
+        return self._model.encode_image(pixel_values)
--- a/server/clip_server/model/openclip_model.py
+++ b/server/clip_server/model/openclip_model.py
@@ -157,10 +157,8 @@ class OpenCLIPModel(CLIPModel):
             return 'ViT-L-14-336'
         return self._model_name.replace('/', '-')
 
-    def encode_text(
-        self, input_ids: 'torch.Tensor', attention_mask: 'torch.Tensor', **kwargs
-    ):
-        return self._model.encode_text(input_ids, attention_mask, **kwargs)
+    def encode_text(self, input_ids: 'torch.Tensor', **kwargs):
+        return self._model.encode_text(input_ids, **kwargs)
 
     def encode_image(self, pixel_values: 'torch.Tensor'):
         return self._model.encode_image(pixel_values)
--- a/server/clip_server/model/openclip_model.py
+++ b/server/clip_server/model/openclip_model.py
@@ -7,7 +7,6 @@
 
 from typing import TYPE_CHECKING
 from copy import deepcopy
-import warnings
 import torch
 
 from clip_server.model.clip_model import CLIPModel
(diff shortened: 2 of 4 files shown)


server/clip_server/model/openclip_model.py
```diff
@@ -20,17 +28,139 @@ class OpenCLIPModel(CLIPModel):
...
+
+                        model.apply(patch_float)
+                        patch_float(model.encode_image)
+                        patch_float(model.encode_text)
+                        model.float()
+
+                    # ensure image_size attr available at consistent location for both jit and non-jit
+                    model.visual.image_size = model.input_resolution.item()
+                if precision == "fp32":
+                    model = model.float()
```
numb3r3, line 137: we can simply use `load_openai_model()` here.

server/clip_server/model/openclip_model.py
```diff
@@ -20,17 +28,139 @@ class OpenCLIPModel(CLIPModel):
...
+            else:
+                if model_name in _MODEL_CONFIGS:
+                    model_cfg = deepcopy(_MODEL_CONFIGS[model_name])
+                else:
+                    raise RuntimeError(f'Model config for {model_name} not found.')
+
+                model = CLIP(**model_cfg)
+
+                if pretrained:
```
numb3r3, line 146: in our case, `pretrained` cannot be None/empty.

server/clip_server/model/openclip_model.py
```diff
@@ -20,17 +28,139 @@ class OpenCLIPModel(CLIPModel):
...
+                    if model_path:
+                        model.load_state_dict(load_state_dict(model_path))
+                    else:
```
numb3r3, line 149: `else` is not necessary?

server/clip_server/model/openclip_model.py
```diff
@@ -20,17 +28,139 @@ class OpenCLIPModel(CLIPModel):
...
+                        raise RuntimeError(
+                            f'Pretrained weights ({pretrained}) not found for model {model_name}.'
+                        )
+
+                model.to(device=torch.device(device))
+                if precision == "fp16":
+                    convert_weights_to_fp16(model)
```
numb3r3, line 156: change the logic, if `device=cuda`, then we use `fp16`

server/clip_server/model/openclip_model.py
```diff
@@ -20,17 +28,139 @@ class OpenCLIPModel(CLIPModel):
...
+                if jit:
+                    model = torch.jit.script(model)
+
+            self._model = model
             self._model_name = model_name
+        else:
```
numb3r3, line 162: we should make sure all of the models are uploaded to our s3 bucket.

numb3r3 (changes requested):

server/clip_server/model/openclip_model.py
shan-mx: can't do it because load_openai_model contains openclip's download method which disables us to download model from our s3

server/clip_server/model/openclip_model.py
ZiniuYu: i uploaded pt models, should be available in any minutes

server/clip_server/model/mclip_model.py
```diff
@@ -74,4 +74,4 @@ def encode_text(
         )
 
     def encode_image(self, pixel_values: torch.Tensor, **kwargs):
-        return self._model.encode_image(pixel_values)
+        return self._model.encode_image(pixel_values, **kwargs)
```
numb3r3, line 77: no, we are sure here `encode_image` only accepts `pixel_values`. Thus, `kwargs` should not be considered.

numb3r3 (changes requested):

server/clip_server/model/openclip_model.py
numb3r3: no, `load_openai_model`  also accept model_path as the parameter.

server/clip_server/model/openclip_model.py
ZiniuYu: update: any hours 😅poor internet

server/clip_server/model/openclip_model.py
```diff
@@ -35,104 +35,10 @@ def __init__(self, name: str, device: str = 'cpu', jit: bool = False, **kwargs):
             pretrained = 'openai'
```
numb3r3, line 35: please double-check the indent space.

server/clip_server/model/openclip_model.py
```diff
@@ -20,28 +28,44 @@ class OpenCLIPModel(CLIPModel):
...
+            if model_name in _MODEL_CONFIGS:
+                model_cfg = deepcopy(_MODEL_CONFIGS[model_name])
+            else:
+                raise RuntimeError(f'Model config for {model_name} not found.')
+
+            model = CLIP(**model_cfg)
+            model.load_state_dict(load_state_dict(model_path))
+            model.to(device=torch.device(device))
+
+            if device == 'cuda':
```
numb3r3, line 52:
```suggestion
            if str(device) == 'cuda':
```

numb3r3 (changes requested):

numb3r3 (approved):
LGTM 👍

PR #783, opened 2022-07-25
fix: add models and md5

ZiniuYu:
This pr adds open-clip model filenames and their corresponding md5
We now host a copy of open-clip models and convert them to onnx runtime


server/clip_server/model/openclip_model.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, name: str, device: str = 'cpu', jit: bool = False, **kwargs):
         if model_url:
             model_path = download_model(model_url, md5sum=md5sum)
             self._model = load_openai_model(model_path, device=device, jit=jit)
-            self._model_name = name
+            self._model_name = name.split('::')[0]
```
ZiniuYu, line 27: @numb3r3 Check this

shan-mx: > @numb3r3 Check this

should be split[0], I had fixed it in my pr

numb3r3: I think the problem caused here can be addressed by another PR. So let's revert this change in this PR.

numb3r3 (changes requested):

numb3r3 (approved):
LGTM

server/clip_server/model/openclip_model.py
ZiniuYu: we need this to pass ci

PR #786, opened 2022-07-28
chore: remove clip_hg

ZiniuYu:
This PR removes the CLIP hugging face executor since hg does not have new models and every other model can be used from open-clip


.github/workflows/cd.yml
```diff
@@ -42,7 +42,6 @@ jobs:
           python -m pip install wheel
           pip install --no-cache-dir "client/[test]"
           pip install --no-cache-dir "server/[onnx]"
-          pip install --no-cache-dir "server/[huggingface]"
```
numb3r3, line 45: `mclip` would also need this dep

server/setup.py
```diff
@@ -58,7 +59,6 @@
         ]
         + (['onnxruntime-gpu>=1.8.0'] if sys.platform != 'darwin' else []),
         'tensorrt': ['nvidia-tensorrt'],
-        'huggingface': ['transformers>=4.16.2'],
```
numb3r3, line 61: change the dependency key to "mclip"

numb3r3 (changes requested):

server/setup.py
ZiniuYu: why put mclip in extra requirements, since it is now part of basic models to support

server/setup.py
```diff
@@ -58,7 +58,7 @@
...
+        'mclip': ['transformers>=4.16.2'],
```
numb3r3, line 61:
```suggestion
        'transformers': ['transformers>=4.16.2'],
```

.github/workflows/ci.yml
```diff
@@ -114,7 +114,7 @@ jobs:
           python -m pip install wheel pytest pytest-cov
           pip install --no-cache-dir "client/[test]"
           pip install --no-cache-dir "server/[onnx]"
-          pip install --no-cache-dir "server/[huggingface]"
+          pip install --no-cache-dir "server/[mclip]"
```
numb3r3, line 117:
```suggestion
          pip install --no-cache-dir "server/[transformers]"
```

.github/workflows/cd.yml
```diff
@@ -42,7 +42,7 @@ jobs:
...
+          pip install --no-cache-dir "server/[mclip]"
```
numb3r3, line 45:
```suggestion
          pip install --no-cache-dir "server/[transformers]"
```

server/setup.py
numb3r3: I changed my mind, let's use `transformers` as the key. `mclip` actually depends on transformers.

numb3r3 (changes requested):

numb3r3 (approved):
LGTM

PR #787, opened 2022-07-28
fix: no allow client to change server batch size

ZiniuYu:
This PR fix the following things:

- [x] the client can no longer change the server batch size by posting through parameters


numb3r3 (approved):
LGTM

jemmyshin:
LGTM

PR #790, opened 2022-08-01
fix: fp16 inference

numb3r3:


server/clip_server/model/model.py
```diff
@@ -0,0 +1,619 @@
...
+
+    def _convert_weights_to_fp16(l):
+        if isinstance(l, (nn.Conv1d, nn.Conv2d, nn.Linear)):
+            l.weight.data = l.weight.data.half()
+            if l.bias is not None:
+                l.bias.data = l.bias.data.half()
+
+        if isinstance(l, nn.MultiheadAttention):
+            for attr in [
+                *[f"{s}_proj_weight" for s in ["in", "q", "k", "v"]],
```
ZiniuYu, line 604: what are 'in' 'q' 'k' 'v'?

numb3r3: refers to query, key, and values in the attention model respectively.

PR #792, opened 2022-08-02
fix: mclip cuda device

numb3r3:


ZiniuYu (approved):

ZiniuYu:
tested and works as desired

PR #795, opened 2022-08-04
feat: encode text first when both text and uri are presented

ZiniuYu:
In previous versions, when a doc contains both text and URI fields, clip_server will treat it as an image doc.
In this pr we encode text first since the text is a [content attribute](https://docarray.jina.ai/fundamentals/document/attribute/#content-attributes) and URI is only a source of content. 
So now when a doc has text and URI fields, we will consider it as a text doc.

### DIFF
--- a/server/clip_server/executors/helper.py
+++ b/server/clip_server/executors/helper.py
@@ -75,9 +75,7 @@ def preproc_text(
 def split_img_txt_da(doc: 'Document', img_da: 'DocumentArray', txt_da: 'DocumentArray'):
     if doc.text:
         txt_da.append(doc)
-    elif doc.blob or (doc.tensor is not None):
-        img_da.append(doc)
-    elif doc.uri:
+    elif doc.blob or (doc.tensor is not None) or doc.uri:
         img_da.append(doc)
 
 
--- a/server/clip_server/executors/helper.py
+++ b/server/clip_server/executors/helper.py
@@ -73,12 +73,12 @@ def preproc_text(
 
 
 def split_img_txt_da(doc: 'Document', img_da: 'DocumentArray', txt_da: 'DocumentArray'):
-    if doc.uri:
-        img_da.append(doc)
+    if doc.text:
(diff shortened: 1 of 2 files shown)


server/clip_server/executors/helper.py
```diff
@@ -73,12 +73,10 @@ def preproc_text(
 
 
 def split_img_txt_da(doc: 'Document', img_da: 'DocumentArray', txt_da: 'DocumentArray'):
-    if doc.uri:
-        img_da.append(doc)
-    elif doc.blob or (doc.tensor is not None):
-        img_da.append(doc)
-    elif doc.text:
+    if doc.text:
```
numb3r3, line 76: add a unit test please

numb3r3 (changes requested):

tests/test_helper.py
```diff
@@ -17,3 +19,42 @@ def test_numpy_softmax(shape, axis):
...
+    'inputs',
+    [
+        (
+            DocumentArray(
+                [
+                    Document(text='hello, world'),
+                    Document(text='goodbye, world'),
+                    Document(
+                        text='hello, world',
+                        uri='https://docarray.jina.ai/_static/favicon.png',
```
numb3r3, line 34: also coverage the docs with `blob` and `tensor`

numb3r3: `tensor = [0, 1, 2, 3]` or `tensor = np.array([0, 1, 2, 3])`

numb3r3 (changes requested):

tests/test_helper.py
ZiniuYu: 🥹

numb3r3:
@ZiniuYu BTW, do we need also to update the client side?

ZiniuYu:
> @ZiniuYu BTW, do we need also to update the client side?

@numb3r3  client happened to handle text first so I think no need to change

numb3r3 (approved):
LGTM

PR #796, opened 2022-08-05
feat: add warning if input is too large

ZiniuYu:
<s>When the doc to be encoded is broken (contains broken URIs, incomplete files, etc.), the server will raise an error and the whole encoding process is thus stopped.
This is especially infeasible for large encoding tasks with thousands or even millions of docs.
In this pr, we improve the user experience by skipping such broken docs and printing warning messages instead.
The whole process is not interrupted.
Users can thus decide whether to proceed with this task or stop.

- [ ] Print warning messages on the server side
- [ ] Print warning messages on the client side when getting incomplete results (i.e. some docs' embeddings are empty)
- [ ] Add test cases</s>

After several discussions with team members, we decide to add a warning at the very beginning of the client to suggest users use moderate-size documents when the input is too large since there is always one more case to consider

### DIFF
--- a/client/clip_client/client.py
+++ b/client/clip_client/client.py
@@ -313,8 +313,8 @@ class Client:
             )
         elif total > 500:
             warnings.warn(
-                'Input size is too large, it is recommended to use smaller input size for better performance. '
-                'Please ensure the inputs are all valid.'
+                'Input size is too large, the process may fail if the input has anything invalid. '
+                'It is recommended to use a smaller input size.'
             )
 
         from docarray.array.mixins.io.pbar import get_pbar
--- a/client/clip_client/client.py
+++ b/client/clip_client/client.py
@@ -109,6 +109,11 @@ class Client:
                 f'content must be an Iterable of [str, Document], try `.encode(["{content}"])` instead'
             )
 
+        if len(content) > 100 * kwargs.get('batch_size', 8):
+            warnings.warn(
+                f'Document size is too large, it is recommended to use smaller document size for better performance'
+            )
+
         self._prepare_streaming(
             not kwargs.get('show_progress'),
             total=len(content) if hasattr(content, '__len__') else None,
--- a/client/clip_client/client.py
+++ b/client/clip_client/client.py
@@ -313,8 +313,7 @@ class Client:
             )
         elif total > 500:
             warnings.warn(
-                'Input size is too large, the process may fail if the input has anything invalid. '
-                'It is recommended to use a smaller input size.'
+                'Please ensure all the inputs are valid, otherwise the request will be aborted.'
             )
 
         from docarray.array.mixins.io.pbar import get_pbar
--- a/client/clip_client/client.py
+++ b/client/clip_client/client.py
@@ -1,6 +1,7 @@
 import mimetypes
 import os
 import time
+import types
 import warnings
 from typing import (
     overload,
@@ -109,7 +110,9 @@ class Client:
                 f'content must be an Iterable of [str, Document], try `.encode(["{content}"])` instead'
             )
 
-        if len(content) > 100 * kwargs.get('batch_size', 8):
+        if not isinstance(content, types.GeneratorType) and len(
+            content
+        ) > 100 * kwargs.get('batch_size', 8):
             warnings.warn(
                 f'Document size is too large, it is recommended to use smaller document size for better performance'
             )
--- a/client/clip_client/client.py
+++ b/client/clip_client/client.py
@@ -110,14 +110,6 @@ class Client:
                 f'content must be an Iterable of [str, Document], try `.encode(["{content}"])` instead'
             )
 
-        if not isinstance(content, types.GeneratorType) and len(
-            content
-        ) > 100 * kwargs.get('batch_size', 8):
-            warnings.warn(
-                f'Document size is too large, it is recommended to use smaller document size for better performance. '
-                'Please ensure the inputs are all valid.'
-            )
-
         self._prepare_streaming(
             not kwargs.get('show_progress'),
             total=len(content) if hasattr(content, '__len__') else None,
@@ -149,7 +141,7 @@ class Client:
     def _unboxed_result(results: 'DocumentArray'):
         if results.embeddings is None:
             raise ValueError(
-                'empty embedding returned from the server. '
+                'Empty embedding returned from the server. '
                 'This often due to a mis-config of the server, '
                 'restarting the server or changing the serving port number often solves the problem'
             )
@@ -317,7 +309,12 @@ class Client:
         if total is None:
             total = 500
             warnings.warn(
-                'the length of the input is unknown, the progressbar would not be accurate.'
+                'The length of the input is unknown, the progressbar would not be accurate.'
+            )
+        elif total > 500:
+            warnings.warn(
+                'Input size is too large, it is recommended to use smaller input size for better performance. '
+                'Please ensure the inputs are all valid.'
             )
 
         from docarray.array.mixins.io.pbar import get_pbar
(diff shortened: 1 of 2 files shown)


ZiniuYu:
I tried to add warning messages to the client at the result gathering steps but changed to (image) preprocessing steps since in this way the process can proceed normally and no repeated warning will be printed

client/clip_client/client.py
```diff
@@ -140,7 +140,7 @@ def _gather_result(self, response, results: 'DocumentArray'):
     def _unboxed_result(results: 'DocumentArray'):
         if results.embeddings is None:
             raise ValueError(
-                'empty embedding returned from the server. '
+                'Empty embedding returned from the server. '
```
numb3r3, line 143: `results.embeddings` would raise ValueError if some document in the result does not have embedding.

client/clip_client/client.py
```diff
@@ -159,17 +159,30 @@ def _iter_doc(self, content) -> Generator['Document', None, None]:
...
-                        tags={'__created_by_CAS__': True}, uri=c
-                    ).load_uri_to_blob()
+                    try:
+                        yield Document(
+                            tags={'__created_by_CAS__': True}, uri=c
+                        ).load_uri_to_blob()
+                    except Exception:
+                        raise ValueError(
+                            f'Failed to load image {c}. Maybe the URI is broken?'
+                        )
```
numb3r3, line 169: why raise a custom exception? why not throw the exception from docarray?

client/clip_client/client.py
```diff
@@ -159,17 +159,30 @@ def _iter_doc(self, content) -> Generator['Document', None, None]:
...
+                    try:
+                        c.load_uri_to_blob()
+                    except Exception:
+                        warnings.warn(
+                            f'Failed to load image doc with ID {c.id}.'
+                            'Maybe the URI is broken?'
+                            'You will get empty embedding for this doc.'
+                        )
+                    finally:
+                        yield c
```
numb3r3, line 185: at the client side, we should raise an exception to stop the encoding.

server/clip_server/executors/helper.py
```diff
@@ -24,20 +25,27 @@ def preproc_image(
...
+            warnings.warn(f'Failed to load image doc with ID {d.id}.')
+            continue
 
-        # recover doc content
-        d.content = content
+        else:
+            tensors_batch.append(preprocess_fn(d.tensor).detach())
+            filtered_da.append(d)
+            # recover doc content
+            d.content = content
```
numb3r3, line 48: just move this block into the `try ... except`

numb3r3 (changes requested):

client/clip_client/client.py
ZiniuYu: It's kinda too long?

client/clip_client/client.py
ZiniuYu: but what about ppl using clip_client who have encoded 98% of a large da and suddenly broke?
if we do not catch this, we can never get empty embeddings

server/clip_server/executors/helper.py
ZiniuYu: ok, `tensors_batch.append(preprocess_fn(d.tensor).detach())
            filtered_da.append(d)` will always work right?

ZiniuYu: ok, `tensors_batch.append(preprocess_fn(d.tensor).detach())
            filtered_da.append(d)` will always work right?

client/clip_client/client.py
ZiniuYu: what about printing a warning (to handle extreme cases like all are broken)?
since this is the last step, the program will exit either way

JoanFM (commented):
I would like to evaluate the perf loss of this. Why dont we have the client say which doc Ids of a batch failed to be encoded?

ZiniuYu:
> I would like to evaluate the perf loss of this. Why dont we have the client say which doc Ids of a batch failed to be encoded?

@JoanFM I added this logic in the first several commits, but the error always occurred at preprocessing steps, which means that we can never reach to the end to generate some report like stuff saying which docs are failed. The client failed even before it start to post docs to the sever and encode. We need to refactor the client logic a bit.

client/clip_client/client.py
numb3r3: exactly, this is what we discussed yesterday. However, pleas also polish the warning message.

client/clip_client/client.py
numb3r3: it's not about the length. It's about the information we can provide to user.

numb3r3: If the exception from docarray is not good enough, then improve the docarray.

client/clip_client/client.py
numb3r3: ah, yes you are right. But please ignore the broken doc, rather than pass to server in `finally`

numb3r3 (changes requested):

client/clip_client/client.py
ZiniuYu: i can but if we simply ignore them, they may disappear in the result. not 100 sure, double checking

client/clip_client/client.py
ZiniuYu: ok i will try gpt3 this time

client/clip_client/client.py
```diff
@@ -109,6 +110,13 @@ def encode(self, content, **kwargs):
                 f'content must be an Iterable of [str, Document], try `.encode(["{content}"])` instead'
             )
 
+        if not isinstance(content, types.GeneratorType) and len(
+            content
+        ) > 100 * kwargs.get('batch_size', 8):
+            warnings.warn(
+                f'Document size is too large, it is recommended to use smaller document size for better performance'
```
numb3r3, line 117: `Document size` is misleading

numb3r3 (changes requested):

client/clip_client/client.py
numb3r3: maybe we can place this logic in `self._prepare_streaming`

numb3r3 (changes requested):

client/clip_client/client.py
ZiniuYu: ur->

client/clip_client/client.py
```diff
@@ -308,7 +309,12 @@ def _prepare_streaming(self, disable, total):
         if total is None:
             total = 500
             warnings.warn(
-                'the length of the input is unknown, the progressbar would not be accurate.'
+                'The length of the input is unknown, the progressbar would not be accurate.'
+            )
+        elif total > 500:
+            warnings.warn(
+                'Input size is too large, it is recommended to use smaller input size for better performance. '
```
numb3r3, line 316: reject, small input size is not for better performance.

ZiniuYu: iknow, im finding a way of saying this without letting users have a wrong feeling that the service is unstable

client/clip_client/client.py
```diff
@@ -308,7 +309,12 @@ def _prepare_streaming(self, disable, total):
         if total is None:
             total = 500
             warnings.warn(
-                'the length of the input is unknown, the progressbar would not be accurate.'
+                'The length of the input is unknown, the progressbar would not be accurate.'
+            )
+        elif total > 500:
+            warnings.warn(
+                'Input size is too large, the process may fail if the input has anything invalid. '
+                'It is recommended to use a smaller input size.'
```
numb3r3, line 317:
```
warnings.warn('please guarantee all the input documents are healthy, '
                          'otherwise the request will be aborted.')
```

numb3r3 (changes requested):

client/clip_client/client.py
ZiniuYu: 🥺

numb3r3 (approved):

PR #804, opened 2022-08-10
feat: do not send blob from server when it is loaded in client

ZiniuYu:
If the user encodes a document with uri and no blob, the current implementation will load the blob in the client and send it to the server.
This pr saves the bandwidth by not sending back blob from the server

I ran some tests to see the benefit we have from this change.

```{python}
from clip_client import Client
from jina import Document, DocumentArray

c = Client('grpc://0.0.0.0:5001')

def data_gen():
    for _ in range(80):
        yield Document(uri='toy.png')

r = c.encode(data_gen(), batch_size=8, show_progress=True)
r.summary()
```

- Server config: default minibatch_size=32, model: Vit-L/14@336px, replicas: 3

- Client config: default batch_size=8, input DocumentArray size=80 with each Document having a URI of an image of size ~1.5MB

- Before: send 90MB, recv 90MB, cost 3-13min (Server hosted on Beijing GPU 3-7min/Berlin GPU 6-13min)

- After: send 90MB, recv 250KB, cost 1-3min

In general, the time varies because of remote network communications, but the recv package sizes substantially drops

### DIFF
--- a/client/clip_client/client.py
+++ b/client/clip_client/client.py
@@ -308,6 +308,10 @@ class Client:
                     ),
                 )
 
+        for c in content:
+            if hasattr(c, 'tags') and c.tags.pop('__loaded_by_CAS__', False):
+                c.pop('blob')
+
         return self._unboxed_result(results)
 
     def _prepare_streaming(self, disable, total):
@@ -457,5 +461,8 @@ class Client:
                         )
                     ),
                 )
+
+        for d in docs:
+            self._reset_rank_doc(d, _source=kwargs.get('source', 'matches'))
 
         return results
--- a/client/clip_client/client.py
+++ b/client/clip_client/client.py
@@ -160,7 +160,8 @@ class Client:
                 _mime = mimetypes.guess_type(c)[0]
                 if _mime and _mime.startswith('image'):
                     yield Document(
-                        tags={'__created_by_CAS__': True}, uri=c
+                        tags={'__created_by_CAS__': True, '__loaded_by_CAS__': True},
+                        uri=c,
                     ).load_uri_to_blob()
                 else:
                     yield Document(tags={'__created_by_CAS__': True}, text=c)
@@ -169,6 +170,7 @@ class Client:
                     yield c
                 elif not c.blob and c.uri:
                     c.load_uri_to_blob()
+                    c.tags['__loaded_by_CAS__'] = True
                     yield c
                 elif c.tensor is not None:
                     yield c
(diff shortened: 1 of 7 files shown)


server/clip_server/executors/helper.py
```diff
@@ -38,6 +38,8 @@ def preproc_image(
 
         # recover doc content
         d.content = content
+        if d.content.tags.pop('__loaded_by_CAS__'):
```
numb3r3, line 41: use `d.pop('content')`

numb3r3 (changes requested):

server/clip_server/executors/helper.py
```diff
@@ -37,7 +37,10 @@ def preproc_image(
         tensors_batch.append(preprocess_fn(d.tensor).detach())
 
         # recover doc content
-        d.content = content
+        if d.tags.pop('__loaded_by_CAS__', None):
```
numb3r3, line 40: `d.tags.pop(...)` will raise exception if the given key does not exist

server/clip_server/executors/helper.py
```diff
@@ -37,7 +37,10 @@ def preproc_image(
...
+            d.pop('tensor')
+        else:
+            d.content = content
```
numb3r3, line 43:
```suggestion
        if '__loaded_by_CAS__' in d.tags:
            d.tags.pop('__loaded_by_CAS__')
            d.pop('tensor')
        else:
            d.content = content
```

numb3r3 (changes requested):

client/clip_client/client.py
```diff
@@ -119,6 +121,11 @@ def encode(self, content, **kwargs):
                 **self._get_post_payload(content, kwargs),
                 on_done=partial(self._gather_result, results=results),
             )
+
+        for c in content:
+            if isinstance(c, Document) and c.tags.pop('__loaded_by_CAS__', False):
```
numb3r3, line 126:
```suggestion
            if hasattr(c, 'tags') and c.tags.pop('__loaded_by_CAS__', False):
```

numb3r3: This is a broken change

numb3r3: The benefits from change above: 
1. reduce the memory footprint;
2. keep the inputs not updated;

numb3r3: The future better way is to support **in-place update** in `encode(docs)`

numb3r3 (changes requested):
add unittest to check server-side

numb3r3 (approved):
LGTM

PR #808, opened 2022-08-23
docs: docs for retrieval

jemmyshin:
add documentation for clip-retrieval


docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,191 @@
...
+│   tags        ('dict',)      1                False             │
+│   text        ('str',)       1                False             │
+│                                                                 │
+╰─────────────────────────────────────────────────────────────────╯
+```
+
+You don't need to call `client.encode()` explicitly since `client.index()` will handle this for you.
+
+
+Similar to `CLIP-as-service`, you need to prepare a YAML config:
```
ZiniuYu, line 54:
```suggestion
Similar to `clip_server`, you need to prepare a YAML config:
```

docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,191 @@
+# Retrieval in CLIP-as-service
+`CLIP-as-service` offers us high quality embeddings. Retrieval is one of the most common use cases for embeddings. Retrieval in CLIP-as-service can support indexing a very large dataset(millions/billions) and querying within several million seconds.
```
ZiniuYu, line 2: milliseconds

docs/user-guides/retrieval.md
numb3r3: the user can directly use the default yaml we created (e.g., `search_flow.yaml` is install via pip as well), e.g., :
```
python -m clip_server search_flow.yaml
```

docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,191 @@
...
+      jtype: CLIPEncoder
+      metas:
+        py_modules:
+          - clip_server.executors.clip_torch
+  - name: indexer
+  	uses:
+  	  jtype: AnnLiteIndexer
+  	  with:
+  	  	dim: 512
+  	  	data_path: workspace
```
numb3r3, line 72: `data_path` parameter will be deprecated since it cannot work with multi-shards.

docs/user-guides/retrieval.md
jemmyshin: good catch

docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,191 @@
...
+              '/delete': 'ALL', '/status': 'ALL'}
+```
+
+| Parameter               | Description                                                                                                                  |
+|-------------------------|------------------------------------------------------------------------------------------------------------------------------|
+| `shards`                  | Number of shardings. |
+| `polling` | Polling strategy for different endpoints.|
+
+
+### Why different polling strategies are needed for different endpoints?
```
numb3r3, line 177: please provide the reference link as well

numb3r3 (changes requested):

docs/user-guides/retrieval.md
numb3r3: `several millionseconds` is not clear. 10 ms, vs 100ms, vs 500ms. As my understanding, 10ms is serveral ms. I don't think we can archive this performance.

numb3r3 (changes requested):

docs/user-guides/retrieval.md
jemmyshin: what reference did you mean?

jemmyshin: ok i see, link for polling strategies

docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,270 @@
+# Retrieval in CLIP-as-service
+`CLIP-as-service` offers us high-quality embeddings. Retrieval is one of the most common use cases for embeddings. Retrieval in `CLIP-as-service` can support indexing a very large dataset (millions/billions) and querying within 50ms, depending on the machine.
```
ZiniuYu, line 2: CAS supports or CAS can index;
depending on the machine configurations

docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,270 @@
...
+
+In order to implement retrieval, we add an extra indexer after the encoder in CLIP-as-service. We use [`AnnLite`](https://github.com/jina-ai/annlite) in this case.
```
ZiniuYu, line 4: We add an 'Annlite' based indexer in CAS

docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,270 @@
+# Retrieval in CLIP-as-service
+`CLIP-as-service` offers us high-quality embeddings. Retrieval is one of the most common use cases for embeddings. Retrieval in `CLIP-as-service` can support indexing a very large dataset (millions/billions) and querying within 50ms, depending on the machine.
+
+In order to implement retrieval, we add an extra indexer after the encoder in CLIP-as-service. We use [`AnnLite`](https://github.com/jina-ai/annlite) in this case. 
+
+
+## Fast search in CLIP-as-service
+Index and search are easy in `CLIP-as-service`:
```
ZiniuYu, line 8: Indexing and searching

docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,270 @@
...
+        'she smiled, with pain',  # text
+        'https://clip-as-service.jina.ai/_static/favicon.png',  # image
+    ]
+)
+
+# search
+client.search(['smile'])
+```
+
+The results will be like this:
```
ZiniuYu, line 27: will look like

docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,270 @@
...
+from clip_client import Client
+
+client = Client('grpc://0.0.0.0:23456')
+
+# index
+r = client.index(
+    [
+        'she smiled, with pain',  # text
+        'https://clip-as-service.jina.ai/_static/favicon.png',  # image
+    ]
```
ZiniuYu, line 20: would it be better to use docarray?
and add one or two more candidates

docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,270 @@
...
+│   Attribute   Data type      #Unique values   Has empty value   │
+│  ─────────────────────────────────────────────────────────────  │
+│   embedding   ('ndarray',)   1                False             │
+│   id          ('str',)       1                False             │
+│   mime_type   ('str',)       1                False             │
+│   tags        ('dict',)      1                False             │
+│   text        ('str',)       1                False             │
+│                                                                 │
+╰─────────────────────────────────────────────────────────────────╯
+```
```
ZiniuYu, line 49: Can you explain a little bit more about this, people who use it for the first time may get confused by this summary

docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,270 @@
...
+          - annlite.executor
+```
+
+| Parameter               | Description                                                                                                                  |
+|-------------------------|------------------------------------------------------------------------------------------------------------------------------|
+| `dim`                  | Dimension of embeddings. Default is 512 since the output dimension of `CLIP` model is 512.|
+
+
+## How to lower memory footprint?
+Sometimes the indexer will use a lot of memory because the HNSW indexer (which is used by `AnnLite`) is stored in memory. The efficient way to reduce memory footprint is dimension reduction. Retrieval in CLIP-as-service use [`Principal component analysis(PCA)`](https://en.wikipedia.org/wiki/Principal_component_analysis#:~:text=Principal%20component%20analysis%20(PCA)%20is,components%20and%20ignoring%20the%20rest.) to achieve this.
```
ZiniuYu, line 89: I would simply say 'because the Annlite indexer is stored in memory' or something like this. People may wonder what is HNSW if there is no explanation. Please also add link to Annlite everywhere

docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,270 @@
...
+
+### Training PCA
+In order to train a PCA model, you need to prepare training data. 
+```python
+from annlite.index import AnnLite
+import numpy as np
+
+index = AnnLite(dim=512, n_components=128)
+index.train(train_data)
+```
```
ZiniuYu, line 99: Put this block after 'how to create `train_data`'

docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,270 @@
...
+      jtype: CLIPEncoder
+      metas:
+        py_modules:
+          - clip_server.executors.clip_torch
+  - name: indexer
+  	uses:
+  	  jtype: AnnLiteIndexer
+  	  with:
+  	  	dim: 512  # input dimension of PCA
+  	  	n_components: 128  # output dimension of PCA
```
ZiniuYu, line 146: A highlight on this line would be better

docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,270 @@
...
+  	  metas:
+  	  	py_modules:
+  	  	  - annlite.executor
+```
+
+### Memory usage before and after PCA
+Here is the comparison of memory usage before and after PCA when indexing 10 million data:
```
ZiniuYu, line 153: Does the machine configuration matters? If yes, please also include it

docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,270 @@
...
+Differences between `ANY` and `ALL`:
+- `ANY`: requests are sent to one of the executors.
+- `ALL`: requests are sent to all executors.
+
+```{figure} images/polling_stratey.png
+:width: 80%
+
+```
+
+Since one data point only needed to be indexed once, there will only be one indexer executor that will handle this data point. Thus, `ANY` is used for `/index`. On the contrary, search operations needed to be handled by all indexer executors since we don't know which executor store the perfectly matched result.
```
ZiniuYu, line 239:
```suggestion
Since one data point only needs to be indexed once, there will only be one indexer executor that will handle this data point. Thus, `ANY` is used for `/index`. On the contrary, search operations needed to be handled by all indexer executors since we don't know which executor stores the perfectly matched result so `ALL` is used.
```

docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,270 @@
...
+
+```{Warning}
+Increasing the number of shardings will definitely alleviate the memory issue, but it will increase the latency since there will be more network connections between different shards.
+```
+
+
+## How to deploy it on the cloud?
```
ZiniuYu, line 246: is this duplicated with [https://clip-as-service.jina.ai/hosting/on-jcloud/](https://clip-as-service.jina.ai/hosting/on-jcloud/)?

ZiniuYu (changes requested):

docs/user-guides/retrieval.md
jemmyshin: This is the example from cas client api document

docs/user-guides/retrieval.md
jemmyshin: yes but this didn't tell users how to send request to flow

ZiniuYu: maybe after you finished the main part, just add a section on that page?

docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,282 @@
...
+            uri='https://clip-as-service.jina.ai/_static/favicon.png'
+        ),  # online image
+    ]
+)
+
+# search
+client.search(['smile'])
+```
+
+The results will look like this:
```
fogx, line 30: What does this image/section tell me as a user? -> Not quite clear to me, but i don't have a lot of experience with Jina yet

docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,282 @@
...
+│   embedding   ('ndarray',)   1                False             │
+│   id          ('str',)       1                False             │
+│   mime_type   ('str',)       1                False             │
+│   tags        ('dict',)      1                False             │
+│   text        ('str',)       1                False             │
+│                                                                 │
+╰─────────────────────────────────────────────────────────────────╯
+```
+
+You don't need to call `client.encode()` explicitly since `client.index()` will handle this for you.
```
fogx, line 54: Is there a reason that i should expect to call `encode` at this position?

docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,282 @@
+# Retrieval in CLIP-as-service
+`CLIP-as-service` offers us high-quality embeddings. Retrieval is one of the most common use cases for embeddings. Retrieval in `CLIP-as-service` support indexing a very large dataset (millions/billions) and querying within 50ms, depending on the machine.
+
+In order to implement retrieval, we add an [`AnnLite`](https://github.com/jina-ai/annlite) indexer executor after the encoder executor in CLIP-as-service.
```
fogx, line 4: It would be nice to give a high-level overview of how the retrieval works/which steps make up the process -> the part from: How Jina eco-system makes cross-modality simple and efficient

docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,282 @@
...
+          - annlite.executor
+```
+
+| Parameter               | Description                                                                                                                  |
+|-------------------------|------------------------------------------------------------------------------------------------------------------------------|
+| `dim`                  | Dimension of embeddings. Default is 512 since the output dimension of `CLIP` model is 512.|
+
+
+## How to lower memory footprint?
+Sometimes the indexer will use a lot of memory because the HNSW indexer (which is used by `AnnLite`) is stored in memory. The efficient way to reduce memory footprint is dimension reduction. Retrieval in CLIP-as-service use [`Principal component analysis(PCA)`](https://en.wikipedia.org/wiki/Principal_component_analysis#:~:text=Principal%20component%20analysis%20(PCA)%20is,components%20and%20ignoring%20the%20rest.) to achieve this.
```
fogx, line 92: Do we know _when_ it will use a lot of memory? Are there any drawbacks to PCA?
Otherwise this seems like a freebie that should always be done -> less memory, same performance

docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,282 @@
+# Retrieval in CLIP-as-service
+`CLIP-as-service` offers us high-quality embeddings. Retrieval is one of the most common use cases for embeddings. Retrieval in `CLIP-as-service` support indexing a very large dataset (millions/billions) and querying within 50ms, depending on the machine.
+
+In order to implement retrieval, we add an [`AnnLite`](https://github.com/jina-ai/annlite) indexer executor after the encoder executor in CLIP-as-service.
+
+
```
fogx, line 6: a table of contents would be nice

docs/user-guides/retrieval.md
jemmyshin: Before you call `client.index()`, you should have embeddings first.

docs/user-guides/retrieval.md
jemmyshin: Good idea,  will do it.

docs/user-guides/retrieval.md
jemmyshin: a table about what?

numb3r3 (changes requested):

docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,292 @@
+# Retrieval in CLIP-as-service
+### Basics of retrieval
+Retrieval is one of the most common use cases for embeddings. Usually retrieval contains two parts: encoding and indexing:
+
+```{figure} images/retreival.png
+:width: 80%
+```
+
+### Multi-modality retreival in `CLIP-as-service`
```
numb3r3, line 9: typo "retreival"

docs/user-guides/retrieval.md
```diff
@@ -0,0 +1,292 @@
...
+  - name: encoder
+    uses:
+      jtype: CLIPEncoder
+      metas:
+        py_modules:
+          - clip_server.executors.clip_torch
+  - name: indexer
+    uses:
+      jtype: AnnLiteIndexer
+      with:
```
numb3r3, line 89: I prefer to place the server flow part before the client example.

numb3r3 (changes requested):

docs/user-guides/retriever.md
```diff
@@ -0,0 +1,275 @@
+# Retrieval in CLIP-as-service
+### Basics of retrieval
+Retrieval is one of the most common use cases for embeddings. Usually retrieval contains two parts: encoding and indexing:
+
+```{figure} images/retreival.png
+:width: 80%
+```
+
+### Multi-modality retreival in `CLIP-as-service`
```
numb3r3, line 9: typo

numb3r3 (changes requested):

docs/user-guides/retrieval.md
fogx: [something like this](https://en.wikipedia.org/wiki/Table_of_contents) :)

jemmyshin: I think we already have one here on the right side of the page: 
Screen Shot 2022-09-07 at 2 47 32 PM

docs/user-guides/retriever.md
```diff
@@ -0,0 +1,326 @@
+# Search API
+
+
+## Basics of CLIP Search
```
ZiniuYu, line 4: I'd recommend remove this title

docs/user-guides/retriever.md
```diff
@@ -0,0 +1,326 @@
...
+```text
+she smiled, with pain defaultdict(<class 'docarray.score.NamedScore'>, {'cosine': {'value': 0.09604912996292114}})
+defaultdict(<class 'docarray.score.NamedScore'>, {'cosine': {'value': 0.7994112372398376}})
+```
+
+You don't need to call `client.encode()` explicitly since `client.index()` will handle this for you.
+
+
+
+## How to lower memory footprint?
```
ZiniuYu, line 151:
```suggestion
## Lower the memory footprint
```

docs/user-guides/retriever.md
```diff
@@ -0,0 +1,326 @@
...
+
+Sometimes the indexer will use a lot of memory because the HNSW indexer (which is used by `AnnLite`) is stored in memory. The efficient way to reduce memory footprint is dimension reduction. Retrieval in CLIP-as-service use [`Principal component analysis(PCA)`](https://en.wikipedia.org/wiki/Principal_component_analysis#:~:text=Principal%20component%20analysis%20(PCA)%20is,components%20and%20ignoring%20the%20rest.) to achieve this.
+
+### Whether PCA is needed in my case?
```
ZiniuYu, line 155: I'd suggest to remove this title

docs/user-guides/retriever.md
```diff
@@ -0,0 +1,326 @@
...
+
+```{Note}
+We use a single machine which has 90GB memory and 20 cores CPU.
+```
+
+```{Tip}
+However, PCA will definitely lead to information losses since we remove some dimensions. And the more dimensions you remove, the more information losses will be. So the best practice will be estimate the memory usage first (if possible, see below) and choose the reasonable dimension after PCA.
+```
+
+## How to deal with a very large dataset?
```
ZiniuYu, line 256:
```suggestion
## Dealing with large dataset
```

docs/user-guides/retriever.md
```diff
@@ -0,0 +1,326 @@
...
+```
+
+Since one data point only needs to be indexed once, there will only be one indexer executor that will handle this data point. Thus, `ANY` is used for `/index`. On the contrary, we use `ALL` in for `/search` since we don't know which executor stores the perfectly matched result, so the search request should be handled by all indexer executors. (The same reason for using `ALL` in `/update`, `/delete`, `/status`)
+
+```{Warning}
+Increasing the number of shardings will definitely alleviate the memory issue, but it will increase the latency since there will be more network connections between different shards.
+```
+
+
+## How to deploy it on the cloud?
```
ZiniuYu, line 311:
```suggestion
## Deploy to JCloud
```

docs/user-guides/retriever.md
```diff
@@ -0,0 +1,326 @@
...
+Deployment can be easily achieved by using [`jcloud`](https://github.com/jina-ai/jcloud) or [`Amazon Kubernetes(EKS) Cluster`](https://aws.amazon.com/eks/). Taking `jcloud` as an example:
```
ZiniuYu, line 312: Use JCloud not \`jcloud\`

docs/user-guides/retriever.md
```diff
@@ -0,0 +1,326 @@
...
+```
+
+| Parameter               | Description                                                                                                                  |
+|-------------------------|------------------------------------------------------------------------------------------------------------------------------|
+| `shards`                  | Number of shardings. |
+| `polling` | Polling strategies for different endpoints.|
+
+Then you can perform exactly the same operations as we do on a single machine.(`/encode`, `/index` and `/search`)
+
+### Why different [polling strategies](https://docs.jina.ai/how-to/scale-out/?highlight=polling#different-polling-strategies) are needed for different endpoints?
```
ZiniuYu, line 293: Different polling strategies for different endpoints

ZiniuYu:
I'm not sure. Should we put the search in Client doc since it is part of the Client?
We should at least mention the search function in Client intros and link to the search docs

ZiniuYu (approved):

PR #812, opened 2022-08-31
chore: improve model not found error msg

ZiniuYu:
We did not list all available models when the user input a wrong model name in torch runtime previously, this pr fix this
this pr also reformat the error msg, it now looks like

image

### DIFF
--- a/server/clip_server/model/clip_model.py
+++ b/server/clip_server/model/clip_model.py
@@ -35,7 +35,9 @@ class CLIPModel(BaseCLIPModel):
 
                 instance = super().__new__(MultilingualCLIPModel)
             else:
-                raise ValueError(f'The CLIP model name=`{name}` is not supported.')
+                raise ValueError(
+                    f'Model {name} not found; available models = {list(_OPENCLIP_MODELS.keys()) + list(_MULTILINGUALCLIP_MODELS.keys())} '
+                )
         else:
             instance = super().__new__(cls)
         return instance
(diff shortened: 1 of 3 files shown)


numb3r3 (changes requested):
What happens on other backend, `ONNX` and `TensorRT`

server/clip_server/model/clip_model.py
```diff
@@ -35,7 +35,9 @@ def __new__(cls, name: str, **kwargs):
 
                 instance = super().__new__(MultilingualCLIPModel)
             else:
-                raise ValueError(f'The CLIP model name=`{name}` is not supported.')
+                raise ValueError(
+                    f'Model {name} not found; available models = {list(_OPENCLIP_MODELS.keys()) + list(_MULTILINGUALCLIP_MODELS.keys())} '
```
numb3r3, line 39: I would prefer to use `CLIP model name="{name}" ...` to highlight the name

ZiniuYu:
> What happens on other backend, `ONNX` and `TensorRT`

the are fine

server/clip_server/model/clip_model.py
numb3r3: And what's more, we can improve the display by using rich table (or other api), e.g., 
```
- RN50
- RN16
- .....
```

numb3r3 (changes requested):

numb3r3 (approved):

PR #814, opened 2022-08-31
feat: allow custom on_done callback function in encoding

ZiniuYu:
This pr allows users to pass a custom on_done callback function in encoding.
If none is provided, it will use the default gather_result callback function and extend the results docarray as usual.

The original gather_result is kept as a callback wrapper to update the progress bar


numb3r3:
Let's first hold this PR until we have a clear vision of the UX.

PR #816, opened 2022-09-08
feat: support clip retrieval

numb3r3:
- support multi-model encoding
- add index and search API


client/clip_client/client.py
```diff
@@ -474,3 +486,262 @@ async def arank(self, docs: Iterable['Document'], **kwargs) -> 'DocumentArray':
...
+            )
+
+        for c in content:
+            if hasattr(c, 'tags') and c.tags.pop('__loaded_by_CAS__', False):
+                c.pop('blob')
+
+        return self._unboxed_result(results)
+
+    @overload
+    async def aindex(
```
jemmyshin, line 560: why do we need `aindex` in the client? The logic of index in the server is not blocked since we use a queue

client/clip_client/client.py
```diff
@@ -1,7 +1,6 @@
 import mimetypes
```
ZiniuYu, line 1: I think we can reuse some codes in this file since its getting quite big

client/clip_client/client.py
ZiniuYu: See tips in this section ->https://clip-as-service.jina.ai/user-guides/client/#async-encoding

client/clip_client/client.py
```diff
@@ -118,9 +117,13 @@ def encode(self, content, **kwargs):
         )
         results = DocumentArray()
         with self._pbar:
+            parameters = kwargs.pop('parameters', None)
```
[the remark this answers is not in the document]
numb3r3: Just want to clean kwargs.

client/clip_client/client.py
numb3r3: yes, exactly. But this is not the response of this PR

scripts/streamlit.py
```diff
@@ -0,0 +1,47 @@
+import numpy as np
+import pandas as pd
+import streamlit as st
+from docarray import DocumentArray, Document
+
+from client.clip_client.client import Client
+
+client = Client('grpc://0.0.0.0:61000')
+st.title('Laion400M retrieval')
```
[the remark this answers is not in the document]
numb3r3: this streamlit script looks like a toy, I will delete it from this PR.

tests/conftest.py
```diff
@@ -56,3 +56,22 @@ def make_trt_flow(port_generator, request):
     f = Flow(port=port_generator()).add(name=request.param, uses=CLIPEncoder)
     with f:
         yield f
+
+
+@pytest.fixture(params=['torch'])
+def make_search_flow(tmpdir, port_generator, request):
```
ZiniuYu, line 62: where do you define this `tmpdir`?

client/clip_client/client.py
ZiniuYu: OK but don't do it too soon 🥲pls wait for my PRs

tests/conftest.py
numb3r3: used as `workspace` in annlite executor

ZiniuYu (approved):

PR #815, opened 2022-09-06
feat: in-place result in clip_client; preserve output order by uid

ZiniuYu:
This PR introduces the following changes to CAS:

- [x] `clip_client` used to return a copy of the input with embeddings and ranking Infos in it. It now updates these into the original input.
- [x] the output order is preserved using uid and is now the same as the input order
      **Notice**: the result is not complete if the ids of input are not unique
- [x] return empty list or docarray for empty inputs
- [x] test cases

Basic tests for memory usage:
```{python}
from clip_client import Client
from docarray import Document, DocumentArray
import psutil

def func():
    c = Client('grpc://0.0.0.0:51000')

    da = DocumentArray()
    for _ in range(8):
        da.append(Document(uri='toy.jpeg'))

    r = c.encode(da)
    return da, r

if __name__ == '__main__':
    mem = []
    for _ in range(10):
        a, b = func()
        mem.append(psutil.Process().memory_info().rss / 1024**2)
    print(mem)
    print(sum(mem)/len(mem))
```

Send 10 batches of images with each batch_size = 8

before (no in-place) in MB: 

> 76.23828125, 76.98046875, 78.0234375, 78.99609375, 79.8125, 79.953125, 73.06640625, 73.42578125, 73.6484375, 74.0703125
> avg 76.421484375

after (in-place) in MB: 

> 73.96875, 71.5703125, 72.72265625, 73.1875, 73.3671875, 73.67578125, 74.2265625, 74.234375, 74.26171875, 72.33203125
> avg 73.3546875

Basic tests for time usage:

```{python}
from clip_client import Client
from docarray import Document, DocumentArray
import time

def func():
    c = Client('grpc://0.0.0.0:51000')

    da = DocumentArray()
    for _ in range(1000):
        da.append(Document(uri='toy.jpeg'))
    c.encode(da, show_progress=True)

if __name__ == '__main__':
    timecost = []
    for _ in range(10):
        t1 = time.time()
        func()
        t2 = time.time()
        timecost.append(t2-t1)
    print(timecost)
    print(sum(timecost)/len(timecost))
```

Time for encoding 1000 images (local machine with CPU):

before (no in-place) in second: 

> 229.51802515983582, 228.4744529724121, 228.77719569206238, 229.18066692352295, 228.6370599269867, 228.76180005073547, 228.8146288394928, 234.90695881843567, 264.61701130867004, 268.92049384117126
> avg 237.0608293533325

after (in-place) in second: 

[...]

PR #817, opened 2022-09-08
feat: add option on server to remove input image content after encoding to save space

ZiniuYu:
Please merge https://github.com/jina-ai/clip-as-service/pull/815 first

This PR adds a `drop_image_content` parameter to `clip_server`. If set, it will remove the input image contents (blob and tensor) after encoding to save space. This is especially beneficial for saving bandwidth.

### DIFF
--- a/server/clip_server/executors/clip_onnx.py
+++ b/server/clip_server/executors/clip_onnx.py
@@ -2,6 +2,7 @@ import os
 import warnings
 from multiprocessing.pool import ThreadPool
 from typing import Optional, Dict
+from functools import partial
 
 import onnxruntime as ort
 from clip_server.executors.helper import (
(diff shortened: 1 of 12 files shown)


server/clip_server/executors/clip_onnx.py
```diff
@@ -25,6 +26,7 @@ def __init__(
         minibatch_size: int = 32,
         access_paths: str = '@r',
         model_path: Optional[str] = None,
+        drop_image_content: bool = False,
```
numb3r3, line 29: `drop_image_content` should not be as init argument. Let's take it as a request parameter only.

numb3r3 (changes requested):

server/clip_server/executors/clip_onnx.py
ZiniuYu: ok make sense

numb3r3:
@ZiniuYu to guarantee the safe conflicts resolve, let's use this PR instead.  https://github.com/jina-ai/clip-as-service/pull/824

PR #818, opened 2022-09-09
docs: enable horizontal scrolling in wide tables

ZiniuYu:
This PR enables scrolling in wide tables like here https://clip-as-service.jina.ai/user-guides/server/#model-support

TBH, I don't know why this modification works 😅. I just don't see Jina core using `sphinx_markdown_tables`. The tutorials I found online are all about customizations in CSS. I guess `sphinx_markdown_tables` has some configurations that disable the scroll bar or set some fixed width.


numb3r3 (approved):
LGTM

PR #830, opened 2022-09-26
feat: more flexible model name by converting to lowercase

ZiniuYu:
This PR converts the model name to lowercase. 
This allows more flexibility when inputting the model name since we get a lot longer model names and cause many typing errors wrt cases.


numb3r3:
I want to hold this PR. Because I cannot see the big benefits of this PR.

PR #836, opened 2022-09-30
chore: use open clip naming convention for model names

ZiniuYu:
This pr adapts the naming convention of open clip to give a consistent list of model names. This can reduce confusion for users.
The S3 storage is also updated.

What is the reason for old names?

- Open clip hosts these models on HF Hub, and these are the names listed on hub.

Why new names?

- We want to give users a consistent experience dealing with [open clip model names](https://github.com/mlfoundations/open_clip#pretrained-model-interface). 

Model affected:

- ViT-B-32::laion2B-s34B-b79K -> ViT-B-32::laion2b-s34b-b79k
- ViT-L-14::laion2B-s32B-b82K -> ViT-L-14::laion2b-s32b-b82k
- ViT-H-14::laion2B-s32B-b79K -> ViT-H-14::laion2b-s32b-b79k
- ViT-g-14::laion2B-s12B-b42K -> ViT-g-14::laion2b-s12b-b42k

### DIFF
--- a/docs/user-guides/server.md
+++ b/docs/user-guides/server.md
@@ -79,7 +79,7 @@ Please also note that **different models give different sizes of output dimensio
 | ViT-B-32::laion2b_e16                 | ✅       | ✅    | ✅        | 512              | 577             | 2.93                | 1.40                 |
 | ViT-B-32::laion400m_e31               | ✅       | ✅    | ✅        | 512              | 577             | 2.93                | 1.40                 |
 | ViT-B-32::laion400m_e32               | ✅       | ✅    | ✅        | 512              | 577             | 2.94                | 1.40                 |
-| ViT-B-32::laion2B-s34B-b79K           | ✅       | ✅    | ❌        | 512              | 577             | 2.94                | 1.40                 |
+| ViT-B-32::laion2b-s34b-b79k           | ✅       | ✅    | ❌        | 512              | 577             | 2.94                | 1.40                 |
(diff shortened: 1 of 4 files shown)


docs/user-guides/server.md
```diff
@@ -79,7 +79,7 @@ Please also note that **different models give different sizes of output dimensio
 | ViT-B-32::laion2b_e16                 | ✅       | ✅    | ✅        | 512              | 577             | 2.93                | 1.40                 |
 | ViT-B-32::laion400m_e31               | ✅       | ✅    | ✅        | 512              | 577             | 2.93                | 1.40                 |
 | ViT-B-32::laion400m_e32               | ✅       | ✅    | ✅        | 512              | 577             | 2.94                | 1.40                 |
-| ViT-B-32::laion2B-s34B-b79K           | ✅       | ✅    | ❌        | 512              | 577             | 2.94                | 1.40                 |
```
numb3r3, line 82: Reject, the normalize is performed by the code, rather than using a normalized name.

numb3r3 (changes requested):

docs/user-guides/server.md
ZiniuYu: this follows https://github.com/mlfoundations/open_clip#pretrained-model-interface

numb3r3:
And what's more, I would suggest normalizing the model name underneath. So that we do not need to update the s3 link if we have the same trouble in future.

numb3r3 (approved):

PR #848, opened 2022-10-28
docs: add instructions for using clip server hosted by jina

ZiniuYu:


docs/hosting/by-jina.md
```diff
@@ -0,0 +1,95 @@
+# Hosted by Jina AI
```
numb3r3, line 1: What's the difference betwee *hostted by Jina AI* and *hosted by jCloud*? I think this title is misleading. Maybe we can just use **# Cloud Hosting**?

numb3r3: I think this page would explain the SaaS product.

numb3r3 (changes requested):

docs/hosting/by-jina.md
ZiniuYu: One is hosted **by** Jina, right now the user can only access it but not control it.
The other is hosted **on** Jcloud, the user has full control of it

numb3r3: But this title does not tell the difference

numb3r3 (approved):
LGTM

PR #849, opened 2022-10-28
feat: allow custom callback in clip_client

ZiniuYu:
`clip_client` by default collects all the results and returns them to users. In this pr, we enable the user to override this behavior by providing custom callbacks: `on_done`, `on_error`, and `on_always`. The usage of custom callbacks follows that of [Jina client](https://docs.jina.ai/fundamentals/client/client/#callbacks). 

We move all results collection logic to a default `on_done` callback, with or without async.
We also move the code that updates the progress bar to a default `on_always` callback, which wraps the user-defined `on_always` callback.
We do not provide a default `on_error` callback.

This PR also fixes a small bug that unboxes empty input for the rank endpoint to unify the behavior of all endpoints.

- [x] All custom callback in `clip_client`
- [x] Add test cases
- [x] Add documentation
- [x] Bug fix: not unboxing empty input for ranking

### DIFF
--- a/client/clip_client/client.py
+++ b/client/clip_client/client.py
@@ -19,17 +19,18 @@ from docarray import DocumentArray
 if TYPE_CHECKING:
     import numpy as np
     from docarray import Document
+    from jina.clients.base import CallbackFnType
 
 
 class Client:
     def __init__(self, server: str, credential: dict = {}, **kwargs):
         """Create a Clip client object that connects to the Clip server.
-        Server scheme is in the format of `scheme://netloc:port`, where
+        Server scheme is in the format of ``scheme://netloc:port``, where
             - scheme: one of grpc, websocket, http, grpcs, websockets, https
             - netloc: the server ip address or hostname
             - port: the public port of the server
         :param server: the server URI
-        :param credential: the credential for authentication {'Authentication': '<token>'}
+        :param credential: the credential for authentication ``{'Authentication': '<token>'}``
         """
         try:
             r = urlparse(server)
@@ -240,6 +241,7 @@ class Client:
         batch_size: Optional[int] = None,
         show_progress: bool = False,
         parameters: Optional[dict] = None,
+        on_done: Optional['CallbackFnType'] = None,
     ) -> 'np.ndarray':
         """Encode images and texts into embeddings where the input is an iterable of raw strings.
         Each image and text must be represented as a string. The following strings are acceptable:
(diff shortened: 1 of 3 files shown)


numb3r3 (changes requested):
Please always document new feature like this.

numb3r3 (changes requested):
And please add unitest

jemmyshin:
please add description for this pr

client/clip_client/client.py
```diff
@@ -221,14 +230,9 @@ def _get_post_payload(self, content, results: 'DocumentArray', kwargs):
...
-    def _unboxed_result(results: 'DocumentArray', unbox: bool = False):
-        if results.embeddings is None:
-            raise ValueError(
-                'Empty embedding returned from the server. '
-                'This often due to a mis-config of the server, '
-                'restarting the server or changing the serving port number often solves the problem'
-            )
-        return results.embeddings if unbox else results
+    def _unboxed_result(results: Optional['DocumentArray'] = None, unbox: bool = False):
+        if results is not None:
```
jemmyshin, line 234: why remove this raise ValueError?

tests/test_client.py
```diff
@@ -27,6 +27,12 @@ async def process(self, docs, **kwargs):
         return results
 
 
+class ErrorExec(Executor):
+    @requests
+    def foo(self, docs, **kwargs):
+        1 / 0
```
jemmyshin, line 33: raise an error here,  do not do like this

tests/test_client.py
```diff
@@ -74,76 +80,173 @@ def test_client_large_input(make_torch_flow):
...
-    assert len(r) == 0
 
-    r = await c.arank(inputs if not callable(inputs) else inputs())
-    if isinstance(inputs, DocumentArray):
-        assert isinstance(r, DocumentArray)
+@pytest.mark.parametrize(
+    'inputs',
+    [
+        [],
+        DocumentArray([]),
```
jemmyshin, line 105: DocumentArray()

jemmyshin (changes requested):

tests/test_client.py
ZiniuYu: hahahaha

client/clip_client/client.py
ZiniuYu: ok i will restore

numb3r3 (approved):
LGTM

PR #851, opened 2022-11-09
fix: install transformers for executor docker images

ZiniuYu:
This pr installs the missing `transformers` package for torch & onnx executor docker images in hubble so that users can use M-CLIP models.

It also updates the hubble executor readme.

### DIFF
--- a/Dockerfiles/cuda.Dockerfile
+++ b/Dockerfiles/cuda.Dockerfile
@@ -3,7 +3,7 @@ ARG CUDA_VERSION=11.4.2
 FROM nvcr.io/nvidia/cuda:${CUDA_VERSION}-cudnn8-runtime-ubuntu20.04
 ENV DEBIAN_FRONTEND=noninteractive
 
-ARG JINA_VERSION=3.7.0
+ARG JINA_VERSION=3.11.0
 ARG BACKEND_TAG=torch
 
 # constant, wont invalidate cache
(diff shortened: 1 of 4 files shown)


Dockerfiles/cuda.Dockerfile
```diff
@@ -27,8 +27,8 @@ COPY . /cas/
 
 WORKDIR /cas
 
-RUN if [ "${BACKEND_TAG}" != "torch" ]; then python3 -m pip install --no-cache-dir "./[${BACKEND_TAG}]" ; fi \
-    && python3 -m pip install --no-cache-dir .
+RUN if [ "${BACKEND_TAG}" != "torch" ]; then python3 -m pip install --no-cache-dir "./[${BACKEND_TAG}]" ;  \
+    else python3 -m pip install --no-cache-dir "./[transformers]" ; fi && python3 -m pip install --no-cache-dir .
```
numb3r3, line 31: Have you verified this line works as we expected?

numb3r3 (changes requested):

Dockerfiles/cuda.Dockerfile
ZiniuYu: I built it locally and it can successfully load M-CLIP models

numb3r3 (approved):
LGTM

PR #852, opened 2022-11-09
feat: support apple silicon

ZiniuYu:
This pr adds support for apple silicon by converting models to `mps` device


numb3r3:
I will first close this PR, and reopen it when we have better idea to support mps device.

PR #854, opened 2022-11-09
fix: increase timeout ready for executor docker images

ZiniuYu:
ViT-g/h and M-CLIP models usually take longer to download and have a high chance to timeout at the starting stage. This pr increases `timeout_ready` to 3000000ms


numb3r3 (approved):
LGTM

PR #856, opened 2022-11-11
chore: allow test to pass even if commit name is not good

ZiniuYu:
Follows https://github.com/jina-ai/jina/pull/5371

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,10 +8,43 @@ jobs:
   commit-lint:
     runs-on: ubuntu-latest
     steps:
+      - name: find the prev warning if exist
+        uses: peter-evans/find-comment@v1
+        id: fc
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          comment-author: "github-actions[bot]"
+          body-includes: "bad commit message"
+      - name: Delete comment if exist
+        if: ${{ steps.fc.outputs.comment-id != 0 }}
+        uses: actions/github-script@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -8,46 +8,10 @@ jobs:
...
-          body: |
-            Thanks for your contribution :heart:
-            :broken_heart: Unfortunately, this PR has one ore more **bad commit messages**, it can not be merged. To fix this problem, please refer to:
-            - [Commit Message Guideline for the First Time Contributor](https://github.com/jina-ai/jina/issues/553)
-            - [Contributing Guideline](https://github.com/jina-ai/jina/blob/master/CONTRIBUTING.md)
-
-            Note, other CI tests will *not* *start* until the commit messages get fixed.
-
-            This message will be deleted automatically when the commit messages get fixed.
-          reaction-type: "eyes"
```
numb3r3, line 50: Is this the expected behavior to drop the warning comments? maybe it is good enough to revise the comment content

numb3r3 (changes requested):

.github/workflows/ci.yml
ZiniuYu: I think it's ok to drop it, since we do `squash and merge`. We only need to make sure the title to meet our criteria.

ZiniuYu: okk

numb3r3 (approved):
LGTM

PR #859, opened 2022-11-15
feat: bump openclip to v2.5.0

OrangeSodahub:
Changes:
- Add `LayernormFp32` (original `Layernorm` handles fp16`);
- Split original `CLIP` to `TextTransformer`, `VisionTransformer` and add `_build_text_tower` `_build_vision_tower`;
- Rearrange modules;


ZiniuYu:
Note: be careful with fp16

numb3r3:
@OrangeSodahub Please make sure the changes will not change the embedding result for each model.

Issue #865, opened 2022-11-22
Installing from readme instructions does not work, Flow is aborted due to ['clip_t'] can not be started.


flamz3d:
create clean python 3.7 env on debian linux

run 
pip install clip-server
run
python -m clip_server

result
```python
           raise FileNotFoundError(f'can not find {path}')                                                                       
       FileNotFoundError: can not find clip_server.executors.clip_torch                                                          
ERROR  Flow@4608 Flow is aborted due to ['clip_t'] can not be started.                                        [11/22/22 13:51:07]
Traceback (most recent call last):
  File "/opt/conda/envs/jina/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/opt/conda/envs/jina/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/opt/conda/envs/jina/lib/python3.7/site-packages/clip_server/__main__.py", line 25, in <module>
    with f:
  File "/opt/conda/envs/jina/lib/python3.7/site-packages/jina/orchestrate/flow/base.py", line 1556, in __enter__
    return self.start()
  File "/opt/conda/envs/jina/lib/python3.7/site-packages/jina/orchestrate/flow/builder.py", line 33, in arg_wrapper
    return func(self, *args, **kwargs)
  File "/opt/conda/envs/jina/lib/python3.7/site-packages/jina/orchestrate/flow/base.py", line 1620, in start
    self._wait_until_all_ready()
  File "/opt/conda/envs/jina/lib/python3.7/site-packages/jina/orchestrate/flow/base.py", line 1719, in _wait_until_all_ready
    raise RuntimeFailToStart
jina.excepts.RuntimeFailToStart
```

jemmyshin:
Can you successfully import this?  @flamz3d 
`import importlib` and then `importlib.import_module('clip_server.executors.clip_torch')`
Here is my output for this:
<img width="926" alt="Screen Shot 2022-11-25 at 11 01 38 AM" src="https://user-images.githubusercontent.com/16580382/203891906-209a75b9-c55b-40b8-bee2-504562b039eb.png">

We found that there is a compatibility issue in python3.7 for `torchvision`. For now, the quick solution is to downgrade the `torchvision` to 0.13.0 (the default version is 0.14.0 if you install clip-server using pip install) and everything should be fine.

RowenaHe:
[...]

PR #866, opened 2022-11-25
fix: torchvision version to avoid compatibility issue

jemmyshin:
This pr resolves the [issue](https://github.com/jina-ai/clip-as-service/issues/865). We fix the `torchvision` version to avoid this.

### DIFF
--- a/server/setup.py
+++ b/server/setup.py
@@ -45,7 +45,7 @@ setup(
         'ftfy',
         'torch',
         'regex',
-        'torchvision',
+        'torchvision==0.13.0',
         'jina>=3.8.0',
         'prometheus-client',
         'open_clip_torch>=1.3.0',
--- a/server/setup.py
+++ b/server/setup.py
@@ -45,7 +45,7 @@ setup(
         'ftfy',
         'torch',
         'regex',
-        'torchvision<=0.13.0' if sys.version_info == (3, 7, 2) else 'torchvision',
+        'torchvision<=0.13.0' if sys.version_info <= (3, 7, 2) else 'torchvision',
         'jina>=3.8.0',
         'prometheus-client',
         'open_clip_torch>=1.3.0',
--- a/server/setup.py
+++ b/server/setup.py
@@ -45,7 +45,7 @@ setup(
         'ftfy',
         'torch',
         'regex',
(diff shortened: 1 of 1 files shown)


server/setup.py
```diff
@@ -45,7 +45,7 @@
         'ftfy',
         'torch',
         'regex',
-        'torchvision',
+        'torchvision<=0.13.0' if sys.version_info == (3, 7, 0) else 'torchvision',
```
ZiniuYu, line 48: what about
```suggestion
        'torchvision==0.13.0' if sys.version_info == (3, 7, 0) else 'torchvision',
```

numb3r3: `'torchvision<=0.13.0' if sys.version_info <= (3, 7, 2) else 'torchvision',`

numb3r3 (changes requested):

server/setup.py
```diff
@@ -45,7 +45,7 @@
         'ftfy',
         'torch',
         'regex',
-        'torchvision',
+        'torchvision<=0.13.0' if sys.version_info <= (3, 7, 2) else 'torchvision',
```
numb3r3, line 48: Please manually check whether this change works as expected.

numb3r3 (changes requested):

server/setup.py
jemmyshin: Screen Shot 2022-11-25 at 3 42 53 PM
Screen Shot 2022-11-25 at 3 42 41 PM

jemmyshin: Different python version: `3.8.15` and `3.7.0`, we have different `torchvision` version

ZiniuYu, line 48:
```suggestion
        'torchvision<=0.13.0' if sys.version_info[:2] <= (3, 7) else 'torchvision',
```

jemmyshin: what's the difference between these two?

ZiniuYu: python 3.7.x: x ∈ [0, 15]

ZiniuYu: https://www.python.org/downloads/

numb3r3 (approved):

PR #868, opened 2022-11-25
feat: add faster tokenizer

OrangeSodahub:
Add a faster tokenizer (original from [rust-tokenizer](https://github.com/guillaume-be/rust-tokenizers)).


server/clip_server/model/fast_tokenizer.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+@lru_cache()
+def default_vocab():
+    return os.path.join(__resources_path__, 'vocab.json')
+
+
+@lru_cache()
+def default_merges():
+    return os.path.join(__resources_path__, 'merges.txt')
```
numb3r3, line 14: Can we generate the `merges` and `vocab` content dynamically from `bpe_simple_vocab_16e6.txt.gz`? It would be a nightmare to maintain two vocabs.

numb3r3 (changes requested):

ZiniuYu:
牛啊
image

server/clip_server/model/fast_tokenizer.py
OrangeSodahub: ok, will do

server/clip_server/model/fast_tokenizer.py
```diff
@@ -0,0 +1,48 @@
...
+        vocab.extend(['<[startoftext>', '<endoftext>'])
+        vocab = dict(zip(vocab, range(len(vocab))))
+        with open(vocab_path, 'w') as f:
+            f.write(json.dumps(vocab))
+    return vocab_path
+
+
+@lru_cache()
+def default_merges():
+    merges_path = os.path.join(__resources_path__, 'merges.txt')
```
numb3r3, line 31: It's too dangerous to change the package data at runtime. Let's keep the installed package static.

server/clip_server/model/fast_tokenizer.py
```diff
@@ -0,0 +1,48 @@
...
+        from simple_tokenizer import bytes_to_unicode
+
+        merges = open(default_merges()).read()
+        merges = merges[1 : 49152 - 256 - 2 + 1]
+        merges = [tuple(merge.split()) for merge in merges]
+        vocab = list(bytes_to_unicode().values())
+        vocab = vocab + [v + '</w>' for v in vocab]
+        for merge in merges:
+            vocab.append(''.join(merge))
+        vocab.extend(['<[startoftext>', '<endoftext>'])
```
numb3r3, line 22: Could you double check the special tokens here? In `simple_tokenizer`, they are using `'<|startoftext|>', '<|endoftext|>'`

server/clip_server/model/fast_tokenizer.py
```diff
@@ -0,0 +1,48 @@
...
+        bpe_path = os.path.join(__resources_path__, 'bpe_simple_vocab_16e6.txt.gz')
+        merges_file = gzip.GzipFile(bpe_path)
+        with open(merges_path, 'w') as f:
+            f.write(merges_file.read().decode('utf-8'))
+    return merges_path
+
+
+class FastTokenizer(object):
+    def __init__(
+        self, vocab_path: str = default_vocab(), merges_path: str = default_merges()
```
numb3r3, line 42: Could you clarity why we need vocab and merge files meanwhile?

numb3r3 (changes requested):

server/clip_server/model/fast_tokenizer.py
OrangeSodahub: To instantiate `PyOpenAiGptTokenizer` we need these two local files

numb3r3:
Since rust-tokenizer is not pip installable. I would close this PR atm.

PR #872, opened 2022-12-07
fix: check dtype when loading models

ZiniuYu:
The dtype is automatically set to `torch.float16` when running on GPU.
This pr fixes dtype checking logics to unblock `torch.float16`


jemmyshin (approved):
LGTM

PR #876, opened 2022-12-08
fix: dynamic convert onnx model to fp16 during start session

OrangeSodahub:
Fix:
- conversion of onnx to fp16 will cover the original model file

Changes:
- move the converter from `__init__()` to `start_session`, using tmp_dir to do the conversion.
- Add test for onnx zip+fp16.

### DIFF
--- a/server/clip_server/model/clip_onnx.py
+++ b/server/clip_server/model/clip_onnx.py
@@ -290,7 +290,7 @@ class CLIPOnnxModel(BaseCLIPModel):
                             )
                         )
                         onnx.save_model(model_fp16, tmp_model_path)
-                    return ort.InferenceSession(tmp_model_path, **kwargs)
+                        model_path = tmp_model_path
             return ort.InferenceSession(model_path, **kwargs)
 
         self._visual_session = _load_session(self._visual_path, 'visual', dtype)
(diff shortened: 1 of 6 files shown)


ZiniuYu (changes requested):

server/clip_server/model/clip_onnx.py
```diff
@@ -279,33 +263,39 @@ def get_model_name(name: str):
 
     def start_sessions(
```
ZiniuYu, line 264: Add a test case for large onnx model zip+fp16

tests/conftest.py
```diff
@@ -40,6 +40,41 @@ def make_flow(port_generator, request):
...
+@pytest.fixture(
+    scope='session',
+    params=[
+        ['onnx', 'ViT-B-32::openai'],
+        ['onnx', 'ViT-H-14::laion2b-s32b-b79k'],
+        ['torch', 'ViT-B-32::openai'],
+        ['onnx_custom', 'ViT-B-32::openai'],
+    ],
+)
+def make_flow_with_large(port_generator, request):
```
ZiniuYu, line 52: Consider move this test to test_model and only test for onnx + fp16 + ViT-H-14::laion2b-s32b-b79k

ZiniuYu (approved):
LGTM

PR #882, opened 2022-12-26
fix: install pytorch cu116 for server docker image

ZiniuYu:
This PR fixes the error in linked issue

### DIFF
--- a/Dockerfiles/server.Dockerfile
+++ b/Dockerfiles/server.Dockerfile
@@ -1,4 +1,4 @@
-ARG CUDA_VERSION=11.6.0
+ARG CUDA_VERSION=11.7.0
 
 FROM nvidia/cuda:${CUDA_VERSION}-devel-ubuntu20.04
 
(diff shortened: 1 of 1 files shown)


jemmyshin (approved):

jemmyshin:
please rebuild the image and reply to the communities.

numb3r3:
@ZiniuYu Does the hub executor also suffer from this issue?

ZiniuYu:
> @ZiniuYu Does the hub executor also suffer from this issue?

I just tested both Torch and ONNX executors, they are all fine

numb3r3:
> I just tested both Torch and ONNX executors, they are all fine

Then, I doubled if this issue comes from cuda? And I would not recommend pinning the cuda version. Our work should work on all stable cuda versions.

numb3r3:
So, please try to make it work with `cuda-11.6.0`, rather than freeze the cuda version.

numb3r3 (approved):
LGTM

PR #883, opened 2022-12-26
chore: bump open-clip-torch to v2.8.0

ZiniuYu:
This PR adapts recent changes introduced in open-clip-torch v2.8+

### DIFF
--- a/server/clip_server/model/model.py
+++ b/server/clip_server/model/model.py
@@ -19,7 +19,7 @@ from clip_server.helper import __cast_dtype__
 from open_clip.transformer import QuickGELU, LayerNorm, LayerNormFp32, Attention
 from open_clip.timm_model import TimmModel
 from open_clip.factory import _MODEL_CONFIGS
-from open_clip.hf_model import PreTrainedTextEncoder
+from open_clip.hf_model import HFTextEncoder
 from open_clip.transformer import ResidualAttentionBlock as _ResidualAttentionBlock
 from open_clip.transformer import Transformer as _Transformer
 from open_clip.transformer import VisionTransformer as _VisionTransformer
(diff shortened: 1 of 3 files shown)


server/clip_server/model/model.py
```diff
@@ -111,6 +120,8 @@ class CLIPVisionCfg:
     patch_size: int = 16
     image_size: Union[Tuple[int, int], int] = 224
     ls_init_value: Optional[float] = None  # layer scale initial value
+    patch_dropout: float = 0.0  # what fraction of patches to dropout during training (0 would mean disabled and no patches dropped) - 0.5 to 0.75 recommended in the paper for optimal results
```
jemmyshin, line 123: why we need dropout during inference? It's not training process

jemmyshin (changes requested):

server/clip_server/model/model.py
ZiniuYu: disabled by default

jemmyshin: Then we can remove this args if we don't need it

tests/test_server.py
```diff
@@ -12,7 +12,7 @@ def test_server_download(tmpdir):
     download_model(
         url='https://docarray.jina.ai/_static/favicon.png',
         target_folder=tmpdir,
-        md5sum='a084999188f4290e2654aec43207ff2e',
+        md5sum='66ea4817d73514888dcf6c7d2b00016d',
```
jemmyshin, line 15: is there any method we can auto-calculate the `md5` in order not to block by this image chaning?

ZiniuYu: no

jemmyshin (approved):
lgtm

tests/test_server.py
jemmyshin: https://github.com/jina-ai/clip-as-service/blob/0b293ec834e80f7335aa625d683904594373a607/server/clip_server/model/pretrained_models.py#L124

PR #894, opened 2023-02-21
fix: onnx package conflict during setup

ZiniuYu:
There is a package conflict with `onnxruntime` and `onnxruntime-gpu` in setup. If we install both packages, we are not able to run onnx using gpu. Installing only `onnxruntime-gpu` enables clip_server to run both on CPU and GPU.

This PR also corrects the instructions on running docker image in different runtimes.

This PR also replaces the broken images in test files

Notes: onnxruntime is fixed at 1.13.1 to pass the test_ranker
nvidia-tensorrt is fixed at 8.4.1.5 to pass trt related tests

### DIFF
--- a/server/setup.py
+++ b/server/setup.py
@@ -53,8 +53,9 @@ setup(
         'onnx': [
             'onnx',
             'onnxmltools',
+            'onnxruntime',
         ]
-        + (['onnxruntime-gpu>=1.8.0'] if sys.platform != 'darwin' else ['onnxruntime']),
+        + (['onnxruntime-gpu>=1.8.0'] if sys.platform != 'darwin' else []),
         'tensorrt': ['nvidia-tensorrt'],
         'transformers': ['transformers>=4.16.2'],
         'search': ['annlite>=0.3.10'],
--- a/server/setup.py
+++ b/server/setup.py
@@ -51,9 +51,9 @@ setup(
     ],
     extras_require={
         'onnx': [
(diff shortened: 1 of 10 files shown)


server/setup.py
```diff
@@ -51,9 +51,9 @@
     ],
     extras_require={
         'onnx': [
-            'onnxruntime',
             'onnx',
             'onnxmltools',
+            'onnxruntime',
```
numb3r3, line 56: As you mentioned, we can always install `onnxruntime-gpu`?

numb3r3 (changes requested):

server/setup.py
ZiniuYu: I've tested on x86 machines with/without GPU, both work fine. The CI failed for unknown reasons

server/setup.py
```diff
@@ -51,12 +51,17 @@
     ],
     extras_require={
         'onnx': [
-            'onnxruntime',
+            'onnxruntime==1.13.1',
```
numb3r3, line 54: I would prefer to use `onnxruntime<=1.13.1`

numb3r3 (changes requested):

server/setup.py
ZiniuYu: let me do some permutations to pass the ci first 😅

numb3r3 (approved):
LGTM

PR #897, opened 2023-03-02
feat: set prefetch in client for traffic control

ZiniuYu:
In this PR, we introduce the prefetch in client for traffic control.
This is especially beneficial to large and time consuming requests which can be unexpectedly be closed be remote server and cause timeout watchdog issue.
This PR also disables uploading progress bar that is not compatible with prefetching 🙁
For more info about client prefetching, please view https://docs.jina.ai/concepts/client/rate-limit/


numb3r3:
Could you give an example of the new progress bar after this PR?

client/clip_client/client.py
```diff
@@ -153,9 +153,9 @@ def _prepare_streaming(self, disable, total):
         os.environ['JINA_GRPC_SEND_BYTES'] = '0'
         os.environ['JINA_GRPC_RECV_BYTES'] = '0'
 
-        self._s_task = self._pbar.add_task(
-            ':arrow_up: Send', total=total, total_size=0, start=False
-        )
+        # self._s_task = self._pbar.add_task(
```
numb3r3, line 156: If this is not really used, please remove them.

numb3r3 (changes requested):

ZiniuYu:
@numb3r3 
image

client/clip_client/client.py
```diff
@@ -334,6 +341,7 @@ def encode(self, content, **kwargs):
                 on_error=on_error,
                 on_always=partial(self._update_pbar, func=on_always),
                 parameters=parameters,
+                prefetch=prefetch,
```
girishc13, line 344: You can just inline the code like: `prefecth=kwargs.get('prefetch', 1000)`.  You can reduce the number of lines if you do this to other params also.

numb3r3:
@ZiniuYu From the example bar, I would suggest changing `Recv` to `Progress`.

PR #903, opened 2023-04-04
updated readme

themantalope:
updated readme to show command for building docker image locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,6 @@
 <br><br><br>
 </p>
 
-
 <p align=center>
 <a href="https://pypi.org/project/clip_server/"><img alt="PyPI" src="https://img.shields.io/pypi/v/clip_server?label=Release&style=flat-square"></a>
 <a href="https://slack.jina.ai"><img src="https://img.shields.io/badge/Slack-3.1k-blueviolet?logo=slack&amp;logoColor=white&style=flat-square"></a>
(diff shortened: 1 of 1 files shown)


numb3r3 (changes requested):
@themantalope Thanks for your interest. Can the guidelines in our current doc not work for you? https://clip-as-service.jina.ai/user-guides/server/#build

numb3r3:
Ok, I see. You want to build `CUDA` image. May I ask you to add the building CLI in our document instead of README?

themantalope:
Hi @numb3r3 

Yeah that's fine. I wanted to use it as an executor in a `jina flow`. The documentation wasn't clear that was the correct way to locally build for the `flow` use case.

numb3r3:
@themantalope For your purpose, you can directly use the containerized  executor from the hub `jinahub+docker://CLIPTorchEncoder` (https://cloud.jina.ai/executor/gzpbl8jh):

```
jtype: Flow
executors:
  - uses: jinahub+docker://CLIPTorchEncoder
```

In the above example, the executor will be a docker container in a flow. 

But you are right, the document is not clear and misleading. Thanks for your pointing out.

themantalope:
@numb3r3 the reason why I wanted to build is because I needed to fork the code and add debug statements. Even on `JINA_LOG_LEVEL=DEBUG` I was getting errors from the `indexer` stating than a `Doc.embedding` was `None`. I was trying to figure out which piece of data it was etc.

PR #911, opened 2023-04-19
fix: fix docarray at v1

ZiniuYu:
This PR aims to prevent the installation of DocArrayV2, as we have encountered some compatibility issues that require further investigation. While this solution is temporary, it will allow us to ensure that our codebase remains stable and free from potential conflicts.

After conducting a thorough investigation of the compatibility of DocArrayV2, we will determine the best course of action moving forward. Until then, we appreciate your patience and understanding as we work to maintain the integrity of our codebase.


numb3r3 (approved):
LGTM

PR #918, opened 2023-06-13
docs: replace free service docs with inference docs

ZiniuYu:


README.md
```diff
@@ -38,18 +38,10 @@ CLIP-as-service is a low-latency high-scalability service for embedding images a
...
-
-```bash 
-jina auth token create <name of PAT> -e <expiration days>
-```
-
-Then, you need to configure the access token in the parameter `credential` of the client in python or set it in the HTTP request header `Authorization` as `<your access token>`.
-
-⚠️ Our demo server `demo-cas.jina.ai` is sunset and no longer available after **15th of Sept 2022**. 
-
+You can access to the hosted CLIP service at [Jina AI's Inference](https://cloud.jina.ai/user/inference) with free credits.
```
jemmyshin, line 41: Can we clarify `free credits` here?

numb3r3 (approved):
LGTM

